PR #19, opened 2018-11-20
better compile error messages WIP/sketch

georgejecook:
NOTE - THIS IS A WIP PR, to show what I'm doing. IT IS NOT TO BE MERGED.

This mechanism looks like it will work fine. I however, have problems/outstanding tasks:

1. extension.ts line 64, whereby, I cannot get the diagnostic to appear unless I use the uri from a document returned from the workbench. A potential workaround is to open the document, without showing it.
1. The code to get the compile errors in the debug session appears wrong - it doesn't return all of the errors from the roku telnet session
1. I'm currently not pulling in the error messages from the debug session; we need to update the regexp to report them - ideally charStart, charEnd, errorMessage and source code. It's all there and diagnostic event handler in extension is already expecting the fields to be set; the regex just needs to do the right thing.

These are all relatively quick to fix. It seems like you concur with the approach. I will also add a command to `Clear Roku Errors` - because we will only be able to clear those on launch/debug, and that might get tedious.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,9 +3,13 @@
         "out": false,
         "node_modules": false,
         ".nyc_output": true,
+        "**/.nyc_output": true,
+        ".nyc_output/**": true,
         ".roku-deploy-staging": false,
         "coverage": true,
         "**/node_modules": true,
-        "**/out": true
+        "**/out": true,
+        "node_modules/**": true,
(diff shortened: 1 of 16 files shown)


georgejecook:
This was just a WIP PR  to facilitate discussion - wasn't intended to be merged.

PR #37, opened 2018-11-23
Compile reporting refactoring signatures better ignores

georgejecook:
This is a grab-bag of features I was developing while working crazy hours this week. It supports:

- Inline error reports on compile fail
- Method signature help
- Improved goto definition
- Refactors how the extension handles the unhandled texts, and introduces the notion of state when initializing.

In addition, also refactors the implementation to better separate classes for future dev.

Please note. All checks pass; but I've been doing this like a mad-man as I needed the features : there's no extra tests at this point.

Saying that - I've got some issues which I'm unsure come from this build or not:
Sometimes there's a Run Time Error; but the debugger doesn't enter debugging mode
I want to make the app exit run state on compile error but don't know how.


TwitchBronBron:
The code coverage percent has decreased. Can you add some unit tests to cover and validate the new functionality that you have added in this PR?

georgejecook:
I can add some tests; but we’re gonna have to horse-trade about coverage :
There’s no way I can get coverage for all of this any time soon, and the
features work and are useful, other’s are already using my branch, such is
their utility. Also, I’m shortly gonna rip out a lot this code and replace
the regex’s with your Brightscript parser so spending time writing tests
now to attain an arbitrary coverage number doesn’t seem like a worthwhile
pursuit imho, and with all due respect. I usually develop with TDD; but I
repurposed code from several places while learning how to write all of
this, so I didn’t benefit from using tests as struts/rigs in this case.

I do concur that some testing needs to happen, especially with the changes
I made to your debugger code to introduce states. The question is: what’s
going to be acceptable to you?


On November 28, 2018 at 7:13:53 AM, Bronley Plumb (notifications@github.com)
wrote:

The code coverage percent has decreased. Can you add some unit tests to
cover and validate the new functionality that you have added in this PR?

—
You are receiving this because you authored the thread.
<https://github.com/TwitchBronBron/vscode-brightscript-language/pull/37#issuecomment-442426266>,
or mute the thread

TwitchBronBron:
I definitely am not expecting you to hit 100% code coverage for your changes. For now, just shoot for writing a test or two that validates the major pieces of functionality, so that we can try to guard against regression issues in the future.

georgejecook:
I'm resubmitting a new PR - it's simply too much work to rebase all of this.

PR #59, opened 2018-12-17
Roku Remote Control

bvisin:
This PR contributes keybindings to send keypresses to the Roku device through Roku's [External Control API](https://sdkdocs.roku.com/display/sdkdoc/External+Control+API#ExternalControlAPI-KeypressKeyValues) using `extension.brightscript.sendRemoteCommand` and passing the key to send as the `args`. 

Example Keybindings:

```
{
	"key": "alt+z",
	"command": "extension.brightscript.toggleRemote"
},
{
	"key": "Backspace",
	"command": "extension.brightscript.sendRemoteCommand",
	"args": "Back",
	"when": "isInRemoteMode && panelFocus && !inDebugRepl && !findWidgetVisible"
},
...
```
Using the above keybindings as an example, you can toggle the Remote on and off by pressing `alt+z`. The remote commands will be sent to the Roku IP that is being used in the last launched debugging session or will prompt you to enter the Roku IP address.

The `when` clause will only send the remote commands if the remote is in an "on" state (`isInRemoteMode`) AND the Panel has focus (`panelFocus`) AND the focus in NOT in the Debug Console REPL (`!inDebugRepl`) AND the Editor Find widget is NOT visible (`!findWidgetVisible`).

You can see an exmaple [keybindings.json file here](https://github.com/bvisin/vscode-brightscript-language/blob/onToggleRemoteMode/example_remote_keybindings.json) to grab (and copy) the 106 keybindings to send the full keyboard and Roku remote buttons.


TwitchBronBron:
Also, I feel like having every interested user copy a giant keybindings file is not the best solution. Can we discuss some alternatives?  I'm still fine with having the option for keybindings (as I see the benefit in assigning keyboard shortcuts to whatever you wish to type...but I think we need something more "built-in" for this feature. Like perhaps a basic popup window with an on-screen keyboard with all known characters, or a textbox and a "send" button or something?

TwitchBronBron:
And great work on this by the way. :D

bvisin:
> One thing I noticed is that when you change the color of the bar at the bottom of vscode, this changes it for every open vscode instance. Is there a way to isolate this setting to only the active window? If I enable roku remote, then close that editor, my other vscode windows bottom bar will stay purple.
> 
> An acceptable workaround is to disable the purple setting on vscode shutdown....if there is such an event?

@TwitchBronBron, I fixed this to only update the local workspace's settings...hopefully that is acceptable, or I can just remove the color bar change...I just liked it because it made it obvious when it was active.

> Also, I feel like having every interested user copy a giant keybindings file is not the best solution. Can we discuss some alternatives? I'm still fine with having the option for keybindings (as I see the benefit in assigning keyboard shortcuts to whatever you wish to type...but I think we need something more "built-in" for this feature. Like perhaps a basic popup window with an on-screen keyboard with all known characters, or a textbox and a "send" button or something?

I agree 100%...I spent a decent amount of time to "capture" the key that triggered the `extension.brightscript.sendRemoteCommand` command, but could not figure that out.  

I tried incorporating all of the keybindings into the extension in the `package.json` but there seems to be an open issue where args can’t be passed from there, https://github.com/Microsoft/vscode/issues/31372

So until that gets fixed I don't have a better idea...your thoughts?

Thanks,
Blake

bvisin:
> Also, I feel like having every interested user copy a giant keybindings file is not the best solution. Can we discuss some alternatives? I'm still fine with having the option for keybindings (as I see the benefit in assigning keyboard shortcuts to whatever you wish to type...but I think we need something more "built-in" for this feature. Like perhaps a basic popup window with an on-screen keyboard with all known characters, or a textbox and a "send" button or something?

I also thought about an on screen remote/text box input, but it was a lot of work to get create a new view in VS Code and we can not put our own views that go where the terminal and debug console are, so I gave up on this idea. I just don't have the time for that right now, and I didn't feel like I would get much value out of it...for my (and @georgejecook's) workflow, the key bindings work brilliantly.

georgejecook:
Second that. I like the fact I can configure the keybindings however I want.

Its unobtrusive too: People don't have to use this feature. They just don't
configure it. No problem imho. So I don't see what harm it would do or why
make something that I've been using non stop since Blake wrote it
unavailable, to wait for another more difficult to implement solution; I'd
not use the other solution anyhow. I just focus on the console log and
press my hotkey. Its marvellous.

I bet many people would use it as it is, as well.  People doing Roku day in
and out don't give a hoot about seeing a picture of a remote control. We
just want to get keypresses on to our devices. Ask in the slack if you
think I'm exaggerating. It a coalface feature.

TwitchBronBron:
> So I don't see what harm it would do or why
make something that I've been using non stop since Blake wrote it
unavailable, to wait for another more difficult to implement solution; I'd
not use the other solution anyhow. I just focus on the console log and
press my hotkey. Its marvellous.

The harm is that once something arrives in this extension, it will need to be supported for a long time. So I only want to put in long-term solutions. This solution works, but it doesn't work out of the box (without a copy-paste action from the user) which means extra setup work for every user who wants this functionality (which let's face it...will probably be most users), and clutters up the kemap file.

I would like to find a solution that solves your desire to `watch the console and press [buttons on your keyboard]`, without needing the manual setup for every single user of this extension. 

This PR also introduces a state (`isInRemoteMode`) to the editor which can be avoided with the right implementation.

Here how I envision this working:

1. Add the standard Roku buttons as bindable hotkey contributions (like `extension.brightscript.sendHomePress`,  `extension.brightscript.sendBackPress`, etc). You can hardcode the actions to specific function calls in code to make them work out of the box without needing users to register the commands themselves. This allows you to send these commands at any time within the editor, and negates the need for `isInRemoteMode`. A keybinding like alt-win (or alt-apple) and then the key should do the trick. 

2. For sending actual text characters, we can leverage the existing debug console text input. This command `vscode.commands.executeCommand('workbench.debug.action.focusRepl');` brings focus to the debug console, and specifically the text input box. So, here's the workflow: 
    - User launches task `extension.brightscript.activateKeyboardInput`, which 
        - calls `vscode.commands.executeCommand('workbench.debug.action.focusRepl');`
        - Tells the extension that the next set of input sent to the debug console should be passed to the roku as "keyboard input" and not evaluated as debug commands. 
    - User types whatever text, and presses the enter key.
    - The extension fires off keyboard events to the roku for every character entered into the textbox.
    - Once the user presses enter, the extension goes back to normal mode.

georgejecook:
The harm is that once something arrives in this extension, it will need to
be supported for a long time. So I only want to put in long-term solutions.

I understand that; I still don’t see the issue of it, if it works, has a
developer supporting it, and doesn’t cause any side-effects, and is
genuinely useful. It’s hands-down my most used feature right now.

This solution works, but it doesn't work out of the box (without a
copy-paste action from the user) which means extra setup work for every
user who wants this functionality (which let's face it...will probably be
most users), and clutters up the kemap file.

I think it’s my choice if I want to clutter up the keymap file, and I
should be allowed to do so if I choose to, so I don’t mind an extension
that has a ton of hooks, that I can use if I choose too. We’re not forcing
anyone to mess up their keymap if they don’t wan to. Even if we come up
with a solution that has a rotating 3d remote and vr hands, or 3d-prints a
remote for you, I’d still want to be able to configure the keymaps how I'd
like, and see this as useful power-user feature


I would like to find a solution that solves your desire to watch the
console and press [buttons on your keyboard], without needing the manual
setup for every single user of this extension.

This PR also introduces a state (isInRemoteMode) to the editor which can be
avoided with the right implementation.

Here how I envision this working:

   1.

   Add the standard Roku buttons as bindable hotkey contributions (like
   extension.brightscript.sendHomePress,
   extension.brightscript.sendBackPress, etc). You can hardcode the actions
   to specific function calls in code to make them work out of the box without
   needing users to register the commands themselves. This allows you to send
   these commands at any time within the editor, and negates the need for
   isInRemoteMode. A keybinding like alt-win (or alt-apple) and then the
   key should do the trick.

It sounds like you’re saying I’ll need to do a chorded key for every
keypress. Is that the case? That’s no good. If not, then it will still stop
me typing code/navigating and searching, also no good. If I’m not
understanding, then let me know what the intent is here; but it sounds like
this won’t fly in for real-world-usage


   1.


   For sending actual text characters, we can leverage the existing debug
   console text input. This command
   vscode.commands.executeCommand('workbench.debug.action.focusRepl'); brings
   focus to the debug console, and specifically the text input box.

I don’t like this: I don’t look at the debug console. It’s difficult to
read and can’t be searched. I look at the output window, which has better
formatting/colors/searchable, etc.


   1.

   So, here's the workflow:
   - User launches task extension.brightscript.activateKeyboardInput, which
         - calls
         vscode.commands.executeCommand('workbench.debug.action.focusRepl');
         - Tells the extension that the next set of input sent to the debug
         console should be passed to the roku as "keyboard input" and
not evaluated
         as debug commands.
      - User types whatever text, and presses the enter key.
      - The extension fires off keyboard events to the roku for every
      character entered into the textbox.
      - Once the user presses enter, the extension goes back to normal mode.

I think if we’re going to have a sensible discussion about this we need to
discuss a real world use case. With all due respect, we’re dogfooding this
feature 6-10 hours every day of the week and the use case you’ve laid out
doesn’t look like any moment of my full time (and more) roku work (I don’t
just start a roku app and type a password, and done). I suggest we use this
real-world use case to plan the feature:

   1. I start the app
   2. I navigate around the app using remote arrow keys, back, ok, star keys
   3. I’m reading log output while I do this
   4. I navigate some more, using the the arrow keys, back, ok,
   5. I navigate to some code files in vscode to double check some log
   output
   6. I navigate some more, back a few times, left , left, right, up
   7. I search my console output for a value
   8. I navigate a bit more, left right, up down
   9. I enter some text in an input field in my roku, by just typing
   10. Press ok a few times
   11. Search my console output
   12. Look for some code files in vscode
   13. Press left, left, left right, down

Currently, I simply activate the roku remote keypress with alt z, and I
have keybindings that only make it work in the log output window when the
search window is not available, and in the console output when not focused
in the repl. I can use hotkeys to jump between my code, the console, the
output window. It’s completely seamless and a delight to use.

I’m sure we can jam out a solution that is less hacky than what we have;
but it has to at least have the same level of comfort and utility. So let’s
focus on the real scenario I’ve provided, and design something that fits
that typical use case.


The ultimate solution will be have a custom window with REPL, roku remote
keypress detection, colored and hyper linkable console output and a preview
coming from a capture card. But the current vscode extension points don’t
support this very well, and I’m not going to invest time to do it using the
html view option, as the LOE doesn’t yield a good enough solution. I’d
never want a custom view, for this as that’s a waste of a left side panel,
imho. I think we have to be realistic about what we want to achieve in
terms of use case and productivity vs vscode limitations.

I hope that I’ve been verbose enough to facilitate a proper discussion
about this now: I’m sure we can find something that works within these
constraints.


—
You are receiving this because you were mentioned.
<https://github.com/TwitchBronBron/vscode-brightscript-language/pull/59#issuecomment-448592842>,
or mute the thread

triwav:
@TwitchBronBron totally understand the wanting to make it not require a user to do a huge key binding file. I think most of the keys like literal letters nobody should have any trouble with but some like play people may want to remap and for sure the key to enable or disable the remote control will definitely want to customize. I would say maybe toggling remote mode could require a keybinding specified by the user. Toggle XML already requires that and while it adds extra effort on the user, it makes the extension not effect global keybindings at all. The others could have defaults set but with the ability to override. Moving the keybindings over to the actual extension itself looks pretty easy as I just looked at an extension doing it https://github.com/Microsoft/vscode-atom-keybindings/blob/master/package.json#L24 

I will side with the others on the effort in adding a UI. If there is enough desire it can be built in. With Roku's own remote tool I only use the interface for selecting what device. The rest is keyboard presses.

TwitchBronBron:
> @triwav As I mentioned above, I tried incorporating all of the keybindings into the extension in the package.json but there seems to be an open (now closed?) issue where args can’t be passed from there, Microsoft/vscode#31372

For the core Roku remote buttons, you shouldn't need to pass in arguments. You would contribute an actual command for each Roku botton in `package.json`.

```json
"commands": [
    {
                "command": "extension.brightscript.pressHomeButton",
                "title": "Press the home button on the Roku remote",
                "category": "Brightscript"
    },
    {
                "command": "extension.brightscript.pressBackButton",
                "title": "Press the back button on the Roku remote",
                "category": "Brightscript"
    }...
```

Then register those commands in `src/BrightScriptCommands.ts`:
```
subscriptions.push(vscode.commands.registerCommand('extension.brightscript.pressHomeButton', (key: string) => {
    this.sendRemoteCommand('Home');
}));

subscriptions.push(vscode.commands.registerCommand('extension.brightscript.pressBackButton', (key: string) => {
    this.sendRemoteCommand('Back');
}));
```

Then contribute keybindings for these in `package.json` (this is untested btw)
```json
"contributes": {
    "keybindings": [{
        "command": "extension.brightscript.pressHomeButton",
        "key": "win+alt+home",
        "mac": "apple+alt+home",
        "when": "editorTextFocus"
    },{
        "command": "extension.brightscript.pressBackButton",
        "key": "win+alt+backspace",
        "mac": "apple+alt+backspace",
        "when": "editorTextFocus"
    }, ...]
}
```

What's nice is that the end user still has the option to override the keybindings to something else.

georgejecook:
I agree for general roku commands.

What happens to typing? i.e. usernames/passwords..


On December 19, 2018 at 3:06:28 PM, Bronley Plumb (notifications@github.com)
wrote:

@triwav <https://github.com/triwav> As I mentioned above, I tried
incorporating all of the keybindings into the extension in the package.json
but there seems to be an open (now closed?) issue where args can’t be
passed from there, Microsoft/vscode#31372
<https://github.com/Microsoft/vscode/issues/31372>

For the core Roku remote buttons, you shouldn't need to pass in arguments.
You would contribute an actual command for each Roku botton in package.json.

"commands": [

    {

                "command": "extension.brightscript.pressHomeButton",

                "title": "Press the home button on the Roku remote",

                "category": "Brightscript"

    },

    {

                "command": "extension.brightscript.pressBackButton",

                "title": "Press the back button on the Roku remote",

                "category": "Brightscript"

    }...

Then register those commands in src/BrightScriptCommands.ts:

subscriptions.push(vscode.commands.registerCommand('extension.brightscript.pressHomeButton',
(key: string) => {

    this.sendRemoteCommand('Home');

}));



subscriptions.push(vscode.commands.registerCommand('extension.brightscript.pressBackButton',
(key: string) => {

    this.sendRemoteCommand('Back');

}));


Then contribute keybindings for these in package.json (this is untested btw)

"contributes": {

    "keybindings": [{

        "command": "extension.brightscript.pressHomeButton",

        "key": "win+alt+home",

        "mac": "apple+alt+home",

        "when": "editorTextFocus"

    },{

        "command": "extension.brightscript.pressBackButton",

        "key": "win+alt+backspace",

        "mac": "apple+alt+backspace",

        "when": "editorTextFocus"

    }, ...]

}

What's nice is that the end user still has the option to override the
keybindings to something else.

—
You are receiving this because you were mentioned.
<https://github.com/TwitchBronBron/vscode-brightscript-language/pull/59#issuecomment-448726870>,
or mute the thread

TwitchBronBron:
vscode already provides an input box popup. What about using it for the keyboard input? 

image

something like `win+alt+k` (popup appears), then type your keys and press enter. 

```
 var stuffUserTyped = await vscode.window.showInputBox({
    placeHolder: 'Press enter to send all typed characters to the roku',
    value: ''
});
```

bvisin:
+1 Will implement!

bvisin:
Okay @TwitchBronBron @georgejecook @triwav  I made some time today to add the popup text input that @TwitchBronBron referenced above.  It works nicely, but I am still trying to figure out how to put the focus back to the `output` panel.  It defaults to the editor after text is entered and you press the enter key.

Check it out and let me know what you guys think.

Thanks!

bvisin:
The focus back to OUTPUT window seems to be a bug https://github.com/Microsoft/vscode/issues/65425

bvisin:
Okay guys, I added the backspace key as the 12th predefined key (cmd+delete)

I also put in a bit of a hacky fix for the focus…In order to bring focus back to the Output panel after inputting text I do
`vscode.commands.executeCommand('workbench.action.output.toggleOutput');` I think this will be acceptable until https://github.com/Microsoft/vscode/issues/65425 gets fixed.

Take a look and let me know what you guys think.  I am pretty happy with the result. Thanks for all the input today.

bvisin:
I changed the command to be `vscode.commands.executeCommand('workbench.action.focusPanel');` which I think _feels_ better than `output.toggleOutput`

TwitchBronBron:
What's the state of this PR. Have you finished all of your changes?

bvisin:
Yes, I believe it's ready and I have made all the changes as you requested.  This PR is waiting on feedback from you and your approval.

PR #64, opened 2019-01-13
Included 'step' as reserved word. Added 'not searchViewletVisible' as condition for keybindings.

miescuderol:
**Included 'step' as reserved word:**
    According to [Roku SDK Docs](https://sdkdocs.roku.com/display/sdkdoc/Program+Statements) `step` is an optional part of `for` loops to specify the increment after each iteration.

 **Added 'not searchViewletVisible' as condition for keybindings**
    `searchViewletVisible` is the name of the search widget if it's part of the panel instead of the sidebar. Without this condition it's practically impossible to use the search panel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -401,73 +401,73 @@
             {
                 "key": "Backspace",
                 "command": "extension.brightscript.pressBackButton",
-                "when": "panelFocus && !inDebugRepl && !findWidgetVisible"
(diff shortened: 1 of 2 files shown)


TwitchBronBron:
Resolves #65 and #66.

TwitchBronBron:
@miescuderol Thanks for your contribution!

PR #70, opened 2019-01-15
`vscode-languageserver` is required for usage as well as development

nsillik:
When running the branch `feature/language-server`, I got errors about the `vscode-languageserver` module not being found. By moving it from `devDependencies` to dependencies`, it seems to be working correctly.


TwitchBronBron:
I was actually just logging in to fix this. Thanks a lot for the fix!

PR #71, opened 2019-01-16
Feature/filtering log output

georgejecook:
Adds improved filtering in the output console, which will help developers see exactly what they need to see during intense debugging sessions.

3 Regex Filters are provided for maximum convenience (all case insensitive):

- LogLevel - example `^\[(info|warn|debug\]`
- Include - example `NameOfSomeInterestingComponent`
- Exclude - example `NameOfSomeNoisyComponent`

Currently, the filters are set by using hotkeys : CMD|WIN+L for level, CMD|WIN+i for include, and CMD|WIN+x for exclude.

The last entered value is remembered - an empty value is presumed to disable the filtering.

I hope someone might build on this to perhaps add a UI/panel/some other better mechanism for managing the filters; but I've been using this for a few days, and it's already been very helpful as it is on 2 huge apps I work on.

### DIFF
--- a/src/LogOutputManager.spec.ts
+++ b/src/LogOutputManager.spec.ts
@@ -1,5 +1,6 @@
 /* tslint:disable:no-unused-expression */
 /* tslint:disable:no-var-requires */
+import { assert } from 'chai';
 import * as sinon from 'sinon';
 let Module = require('module');
 
@@ -16,7 +17,8 @@ Module.prototype.require = function hijacked(file) {
     }
 };
 
-import { LogOutputManager } from './LogOutputManager';
+import { LogLine, LogOutputManager } from './LogOutputManager';
+const itParam = require('mocha-param');
 
 describe('LogOutputManager ', () => {
     let logOutputManagerMock;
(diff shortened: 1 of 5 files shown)


src/LogOutputManager.spec.ts
```diff
@@ -156,4 +170,102 @@ describe('LogOutputManager ', () => {
...
+                const logLine = new LogLine(params.text, false);
+                logOutputManager.setIncludeFilter(params.includeFilter);
+                logOutputManager.setExcludeFilter(params.excludeFilter);
+                logOutputManager.setLevelFilter(params.levelFilter);
+                assert.equal(logOutputManager.matchesFilter(logLine), params.expected);
+            });
+        });
+        describe('non-must include items false scenarios', () => {
+            let mustIncludeParams = [
+          { text: 'INFO test1 INCLUDE', levelFilter: 'DEBUG', includeFilter: null, excludeFilter: null, expected: false },
```
georgejecook: Ironically, the syntax of my mocha inspired unit testing framework, is _way_ more readable than the at-hand options I could find for mocha parameterized testing.. Wow.

PR #75, opened 2019-01-18
Add hyperlinks to code files in BrightScript output channel

### LINKED ISSUE #73: Console output hyperlinking to code files
Would be beyond awesome to have a mechanism for clicking on log lines (or even just embedded filenames) so that we go straight to the line that logged it.

Individual developers (or logging frameworks) could then embed the links to files in the output.

TwitchBronBron:
Fixes #73

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "node"
+  - "11.4.0"
 before_script:
   - npm install
 script:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
  - Added ability to click on `pkg:/` links in BrightScript output window to open that file at the specified line number.
+ - Fixed bug that was not showing an error message when a deployment failed.
 
 ## [1.6.0] - 2019-01-15
  - Prevent adding entry breakpoint when `stopOnEntry` is `false` in launch.json.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## [Unreleased]
+ - Added ability to click on `pkg:/` links in BrightScript output window to open that file at the specified line number.
+
 ## [1.6.0] - 2019-01-15
  - Prevent adding entry breakpoint when `stopOnEntry` is `false` in launch.json.
  - Upgraded to roku-deploy@2.0.0-beta2 which fixes some regression issues introduced in 1.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## [Unreleased]
  - Added ability to click on `pkg:/` links in BrightScript output window to open that file at the specified line number.
- - Fixed bug that was not showing an error message when a deployment failed.
+
+## [1.6.1]
+ - Fixed bug where the debugger would hang indefinitely on certain deployment errors.
 
 ## [1.6.0] - 2019-01-15
  - Prevent adding entry breakpoint when `stopOnEntry` is `false` in launch.json.
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ A VSCode extension to support Roku's BrightScript language.
 - Method signature help (open bracket, or APPLE/Ctrl + SHIFT + SPACE)
 - Roku remote control from keyboard ([click here](#Roku-Remote-Control) for for more information)
 - Brightscript output log (which is searchable and can be colorized with a plugin like [IBM.output-colorizer](https://marketplace.visualstudio.com/items?itemName=IBM.output-colorizer)
+- Navigate to source files referenced as `pkg:/` paths from output log
 - Marking the output log (CTRL+L)
 - Clearing the output log (CTRL+K), which also clears the mark indexes
 - Filtering the output log - 3 filters are available:
--- a/README.md
+++ b/README.md
@@ -189,9 +189,5 @@ View our [developer guidelines](https://github.com/TwitchBronBron/vscode-brights
 
 You can also chat with us [on slack](http://tiny.cc/nrdf0y). (We're in the #vscode-bs-lang-ext channel).
 
-## Known Issues
-
-Click [here](https://github.com/TwitchBronBron/vscode-brightscript-language/issues) to see the list of known issues.
-
 ## Changelog
 Click [here](https://github.com/TwitchBronBron/vscode-brightscript-language/blob/master/CHANGELOG.md) to see the changelog.
--- a/package-lock.json
+++ b/package-lock.json
@@ -4776,9 +4776,9 @@
             }
         },
         "roku-deploy": {
-            "version": "2.0.0-beta2",
-            "resolved": "https://registry.npmjs.org/roku-deploy/-/roku-deploy-2.0.0-beta2.tgz",
-            "integrity": "sha512-bQB4AGM4Zqf39Z2BmdtSdCshn0+iBPMY2+GHvsp8gPVz1nqVzHLdoezrNIl5yHzJoRSNRc2BSe2hdwom3v8F0w==",
+            "version": "2.0.0-beta4",
+            "resolved": "https://registry.npmjs.org/roku-deploy/-/roku-deploy-2.0.0-beta4.tgz",
+            "integrity": "sha512-VMVvs9QkYfyp9KHrVGrmjgWDrpza5QGyONMu4yznKhxbS7bEF1O4bzs5ltiaN+shxIhPU6+8vDmKw2iI0qqS4w==",
             "requires": {
                 "archiver": "^3.0.0",
                 "fs-extra": "^5.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "brightscript",
-    "version": "1.6.0",
+    "version": "1.6.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -563,7 +563,7 @@
                 },
                 "readable-stream": {
                     "version": "2.3.6",
-                    "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+                    "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
                     "integrity": "sha512-tQtKA9WIAhBF3+VLAseyMqZeBjW0AHJoxOtYqSUZNJxauErmLbVm2FW1y+J/YA9dUrAC39ITejlZWhVIwawkKw==",
                     "requires": {
                         "core-util-is": "~1.0.0",
@@ -5976,7 +5976,7 @@
         },
         "yargs": {
             "version": "1.2.6",
-            "resolved": "https://registry.npmjs.org/yargs/-/yargs-1.2.6.tgz",
+            "resolved": "http://registry.npmjs.org/yargs/-/yargs-1.2.6.tgz",
             "integrity": "sha1-nHtKgv1dWVsr8Xq23MQxNUMv40s=",
             "requires": {
                 "minimist": "^0.1.0"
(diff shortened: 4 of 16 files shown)


TwitchBronBron:
@georgejecook I'd love your feedback on this feature before I merge it into master.

georgejecook:
gonna give this a spin with my rLog experimental branch

georgejecook:
Okay - I've got a preprocessor that will work with this. I would _love_ to do the following, and I think it's a protocol that will give max flexibility to everyone. tell me what you think..

protocol is pkg:/path/file.brs(line:col:alternateText)

and if I print the following to telnet

`pkg:/components/KeyLogTester.brs(24:0:   23) [INFO] retriving videos for category id 23`

This will end up in the debug log as `   23 [INFO] retriving videos for category id 23` where `   23` will be the hyperlink, and will resolve to `pkg:/components/KeyLogTester.brs`

I've had a brief look at your impl, and I believe that it will be possible to do this. I'll be pleased to provide the implementation.

The ask is that it means those of us providing log frameworks, or just any developer, with their own logging mechanism, can pretty print useful hyperlinks into their output without messing it up with huge paths, or can choose a condensed format like 

`[INFO] retriving videos for category id 23 @pkg:/components/KeyLogTester.brs(24:0:KeyLogTester.brs(24)) `

I know this is an unusual feature ask; but given vscode's limited output options, it'd be a really useful thing to have.

georgejecook:
and as it goes, in my logging framework, I'd allow devs to choose either format (I'd favour the latter, personally) - dunno about you - sometimes just seeing a file:line number is enough to know 🤦‍♂️  "dammit that bug again"

TwitchBronBron:
I'd like to land this PR first, so we have workable pkg links ASAP for everyone. **Are you ok with that as a first step**? 

### My thoughts on your proposal

Roku itself spits out the standard `pkg:/components/KeyLogTester.brs(24)` format, so by pushing the responsibility of `alternateText` off to a preprocessor, we would end up with a "mostly good" solution that still left any roku-generated pkg lines unshortened. 

**I think we can handle the alternateText entirely from within in the extension**, which keeps this feature the most flexible, and would work even for those devs without a preprocessor (for any pkg paths sent by roku, like error lines and such). 

Most of the magic would happen inside the LogOutputManager. Since the LogOutputManager and DocumentLinkProvider are both operating on the same log output, we could map a line index with the pkg path from that line, allowing us to pretty much display whatever we want in the log and make it linkable (i.e. no need for arbitrary unique numbers like `#1` or the like).

| | |
|--|--|
|Full| `pkg:/components/KeyLogTester.brs(24:0) [INFO] retriving videos for category id 23`|
|Filename| `KeyLogtester.brs(24) [INFO] retriving videos for category id 23`|
|Short |`#1 [INFO] retriving videos for category id 23`|
|Hidden| [INFO] retriving videos for category id 23`|

As far as column numbers, I couldn't figure out a way to construct a Uri that supported column numbers. However, I have updated this branch to at least recognize them so both (21) and (21:8) match to line 21. 

My goal is to make it the most flexible as possible, while still also meeting your needs. From what I can tell, the alternate text was just a uniquely generated number...which we can absolutely do in the extension instead). How would you feel about this solution?

georgejecook:
I’m trying out your branch.. the issue I have is that I lose _all_
coloring.. :( - that would make it unusable for everybody (I think nearly
everyone uses the IBM output colorizer
https://github.com/IBM-Cloud/vscode-log-output-colorizer)

I’ve thought for the longest time that your extension should manage that
btw - you own the log output after all; but I don’t know if it’s possible
to colorize a specific log output - I’m guessing so, as some of the log
outputs are getting colorized still; just not the debug ones.

As for link formats: Totally agree about not worrying about it now. It was
my intention that any further text substitution would happen in another PR;

On that note, your notion of 4 link options works great for me - if we had
a means of knowing how to configure them (or would that be a vscode
setting)?


On January 18, 2019 at 11:31:22 PM, Bronley Plumb (notifications@github.com)
wrote:

I'd like to land this PR first, so we have workable pkg links ASAP for
everyone. *Are you ok with that as a first step*?
My thoughts on your proposal

Roku itself spits out the standard pkg:/components/KeyLogTester.brs(24)
format, so by pushing the responsibility of alternateText off to a
preprocessor, we would end up with a "mostly good" solution that still left
any roku-generated pkg lines unshortened.

*I think we can handle the alternateText entirely from within in the
extension*, which keeps this feature the most flexible, and would work even
for those devs without a preprocessor (for any pkg paths sent by roku, like
error lines and such).

Most of the magic would happen inside the LogOutputManager. Since the
LogOutputManager and DocumentLinkProvider are both operating on the same
log output, we could map a line index with the pkg path from that line,
allowing us to pretty much display whatever we want in the log and make it
linkable (i.e. no need for arbitrary unique numbers like #1 or the like).
Full pkg:/components/KeyLogTester.brs(24:0) [INFO] retriving videos for
category id 23
Filename KeyLogtester.brs(24) [INFO] retriving videos for category id 23
Short #1 [INFO] retriving videos for category id 23
Hidden [INFO] retriving videos for category id 23`

As far as column numbers, I couldn't figure out a way to construct a Uri
that supported column numbers. However, I have updated this branch to at
least recognize them so both (21) and (21:8) match to line 21.

My goal is to make it the most flexible as possible, while still also
meeting your needs. From what I can tell, the alternate text was just a
uniquely generated number...which we can absolutely do in the extension
instead). How would you feel about this solution?

—
You are receiving this because you were mentioned.
<https://github.com/TwitchBronBron/vscode-brightscript-language/pull/75#issuecomment-455748718>,
or mute the thread

georgejecook:
Links don’t work for me either.

```
pkg:/components/controllers/ContentController/ContentController.brs(673)
*[METHOD] GridController.toggleLoadingIndicator, false reason : onLoadItem
got content
pkg:/components/controllers/GridController/GridController.brs(227)
*[METHOD] LoadingIndicator.onControlChange
pkg:/components/lib/LoadingIndicator/LoadingIndicator.brs(175)
*[METHOD] LoadingIndicator.onFadeAnimationStateChange
pkg:/components/lib/LoadingIndicator/LoadingIndicator.brs(193)
*[METHOD] LoadingIndicator.onFadeAnimationStateChange
pkg:/components/lib/LoadingIndicator/LoadingIndicator.brs(193)
*[METHOD] LoadingIndicator.stopAnimation
pkg:/components/lib/LoadingIndicator/LoadingIndicator.brs(281)
```


I presume because it’s not using the mapping from my config file..
i.e.

```
      "debugRootDir":"${workspaceFolder}/templates/default”
```

Those files are in `${workspaceFolder}/templates/default/components…`

On January 19, 2019 at 3:09:08 PM, George Cook (george@tantawowa.com) wrote:

I’m trying out your branch.. the issue I have is that I lose _all_
coloring.. :( - that would make it unusable for everybody (I think nearly
everyone uses the IBM output colorizer
https://github.com/IBM-Cloud/vscode-log-output-colorizer)

I’ve thought for the longest time that your extension should manage that
btw - you own the log output after all; but I don’t know if it’s possible
to colorize a specific log output - I’m guessing so, as some of the log
outputs are getting colorized still; just not the debug ones.

As for link formats: Totally agree about not worrying about it now. It was
my intention that any further text substitution would happen in another PR;

On that note, your notion of 4 link options works great for me - if we had
a means of knowing how to configure them (or would that be a vscode
setting)?


On January 18, 2019 at 11:31:22 PM, Bronley Plumb (notifications@github.com)
wrote:

I'd like to land this PR first, so we have workable pkg links ASAP for
everyone. *Are you ok with that as a first step*?
My thoughts on your proposal

Roku itself spits out the standard pkg:/components/KeyLogTester.brs(24)
format, so by pushing the responsibility of alternateText off to a
preprocessor, we would end up with a "mostly good" solution that still left
any roku-generated pkg lines unshortened.

*I think we can handle the alternateText entirely from within in the
extension*, which keeps this feature the most flexible, and would work even
for those devs without a preprocessor (for any pkg paths sent by roku, like
error lines and such).

Most of the magic would happen inside the LogOutputManager. Since the
LogOutputManager and DocumentLinkProvider are both operating on the same
log output, we could map a line index with the pkg path from that line,
allowing us to pretty much display whatever we want in the log and make it
linkable (i.e. no need for arbitrary unique numbers like #1 or the like).
Full pkg:/components/KeyLogTester.brs(24:0) [INFO] retriving videos for
category id 23
Filename KeyLogtester.brs(24) [INFO] retriving videos for category id 23
Short #1 [INFO] retriving videos for category id 23
Hidden [INFO] retriving videos for category id 23`

As far as column numbers, I couldn't figure out a way to construct a Uri
that supported column numbers. However, I have updated this branch to at
least recognize them so both (21) and (21:8) match to line 21.

My goal is to make it the most flexible as possible, while still also
meeting your needs. From what I can tell, the alternate text was just a
uniquely generated number...which we can absolutely do in the extension
instead). How would you feel about this solution?

—
You are receiving this because you were mentioned.
<https://github.com/TwitchBronBron/vscode-brightscript-language/pull/75#issuecomment-455748718>,
or mute the thread

georgejecook:
the issue with the log colors going, is presumably that we're capturing the whole log in ```            {
                "id": "brightscript-log",
                "mimetypes": [
                    "text/x-code-output"
                ]
            }```

I'm worried that's never going to be sufficiently constrained to not interfere with any other extensions, as per this issue https://github.com/Microsoft/vscode/issues/586#issuecomment-248184108

Unfortunately Microsoft are not going to improve this situation any time soon. 

There's very few options, these are the 2 that spring to mind:
1. create a log ouput language type for brightscript, and capture the behavior for all log output. This would require colorizing, and could quite possibly irritate user's who don't just use vscode for brightscript dev.
2. Assume that most vscode users are going to have the ibm extension installed (i.e. https://github.com/IBM-Cloud/vscode-log-output-colorizer), and bind the link provider (via `registerDocumentLinkProvider`), to their language definition (i.e. id: `Log`, as per https://github.com/IBM-Cloud/vscode-log-output-colorizer/blob/26eb53df4c8a53b5d8f8dfa760137967bc56faf9/package.json#L28)

I favour option 2 currently - I think until Microsoft improve the situation, it's hard to imagine that an extension with 166k and growing installs, isn't going to be somewhat standard - and at least the id Log is somewhat generic. The good news is, that there's no install dependency required. I just verified it by modifying package.json so that the log language definition is identical to that of the imb extension. I.e.

```        "languages": [
            {
                "id": "Log",
                "aliases": [
                    "log"
                ],
                "extensions": [
                    ".log"
                ],
                "configuration": "./log.configuration.json",
                "mimetypes": [
                    "log",
                    "text/log",
                    "text/x-log",
                    "text/x-code-output",
                    "x-code-output"
                ]
            }
        ],
```

And I changed the log registration in `extension.ts` to be: 
```vscode.languages.registerDocumentLinkProvider({ language: 'Log' }, docLinkProvider);```

And it works totally fine - colouring from IBM functions perfectly, as do the hyperlinks themselves.

I hope you can follow this - pm me if not; but to me it looks like you can readily merge your useful branch here, without fear of degrading the experience for those who use multiple extensions, by making the 2 changes I outlined here. I think the only thing missing for those of us with more advanced build-tool-chains, is that the config setting `"debugRootDir"`, be employed when resolving the roku 'pkg' hyperlinks.

TwitchBronBron:
I'm cool with option 2. I'm currently working on fixing the debugRootDir issue now, and I'll change this as well and then have you test it one more time to make sure everything is working properly.

georgejecook:
this is great news. I'm really excited about getting this out.. yes, in part because I can show off rLog; but also - coz rLog can show off your extension :D

this is going to be a great feature imo, and save people a lot of RSI and free up some brain cells (you'll think I'm joking about that, till you're an old fogey like me :D )

TwitchBronBron:
I just added the following fixes:
 - Copied the `Log` language type from the colorizer project and changed link code to use that language.
 - Added support for debugRootDir. 

@georgejecook can you pull latest code and test it again? The colorizer worked for me, as well as the `debugRootDir` fix. 

I'm not exactly sure what the problem is when you say `"Links don’t work for me either"`, so I'm just going to explain something with the hopes that it might be related to your issue.  When using `debugRootDir`, every file in `debugRootDir` and `rootDir` need to have the *exact same location* in both folders. If your build process changes the locations of files between `debugRootDir` and `rootDir` (like moving `src/lib.brs` to `dest/some_subfolder/lib.brs`) there's not much this extension can do about that, as it doesn't have any way to know the file got moved. 

It DOES, however, support using the `{src;dest}` format in the `files` array, so if you are indeed moving files to different subfolders, you could let the extension do the file moving, then it can be smart enough to map your `dest` files back to `src` files. 

Let me know how it goes!

Issue #78, opened 2019-01-19
Read environment variables from external file for debuggin


bvisin:
I would like to have a launch config file like:
```{
    "version": "0.2.0",
    "configurations": [

        {
            "type": "brightscript",
            "request": "launch",
            "name": "Roku Launch",
            "envFile": "${workspaceFolder}/.env",
            "stopOnEntry": false,
            "host": "${env:ROKU_DEV_TARGET}",
            "password": "${env:ROKU_DEVPASSWORD}",
            "rootDir": "${workspaceFolder}/src",
            "consoleOutput": "normal",
            "outDir": "${workspaceFolder}/out",
            "files":[
                "manifest",
                "source/**/*.*",
                "components/**/*.*",
                "images/**/*.*",
                "fonts/**/*.*",
                "csfake/**/*.*"
            ],

        }
    ]
}```

`"envFile": "${workspaceFolder}/.env",`

I think VS Code does this for node debugging: https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_load-environment-variables-from-external-file-node

https://github.com/Microsoft/vscode/issues/15964

https://github.com/Microsoft/vscode/issues/14523

But I would like to add this functionality to the BS debugger as well.

**Comment by bvisin** [comment]
@TwitchBronBron Pulled down the branch and works perfect! Thanks for jumping on this so quickly!  Really appreciate the addition.  I'll continue to use the branch for a while to see if I notice any abnormalities, but I would say this is a great feature that should be included in the next release.

**Comment by bvisin** [comment]
@TwitchBronBron I am seeing an issue with the remote not working as the remote pulls the IP from the launch file...

```send http://${env:ROKU_DEV_TARGET}:8060/keypress/Down
BrightScriptCommands.ts:109
rejected promise not handled within 1 second
extensionHostProcess.js:768
```

So I think we can easily detect `${env:VAR}` and replace that with the env variable...not sure how that all works with different processes as I would like to be loading that in from the .env file but I will try and look into it when I get some time.

bvisin:
Awesome @TwitchBronBron Thank you!

PR #81, opened 2019-01-24
Restructured the resolveDebugConfiguration to save host AFTER process…

bvisin:
…ing .env


bvisin:
This resolves the remote control issue while using .env files reported in https://github.com/TwitchBronBron/vscode-brightscript-language/issues/78

PR #100, opened 2019-02-27
Mechanism to lock alternate text in the logoutput window, while still having links

georgejecook:
# Overview 

This addresses a need I mentioned in our discusison in https://github.com/TwitchBronBron/vscode-brightscript-language/pull/75#issuecomment-455748718

I essentially want to make it possible for any text to become a hyperlink. By looking at the previous discussion, I don't know how that would've worked without it making the whole line a hyperlink, which is really unsightly.

I've opted here to keep the hyperlinks as were, so _any_ and _all_ *pkg:/file.brs(line)* hyperlinks are  generated.

In addition, if log output is detected in the format [#PKG:/#TEXT#PKG_PATH#], then that token is replaced with TEXT, and a custom link is created in the `LogDocumentLinkProvider` to make TEXT a link to PKG_PATH.

This allows anyone to turn any arbitrary text into a link in the log output, which reads way better. Here's a couple of examples : https://imgur.com/a/nDEJrxH

# Changes

- adds notion of custom links to the LogDocumentLinkProvider
- adds additional mock apis to vscode mock
- passes in LogDocumentLinkProvider into LogOutputManager, so it can add and reset custom links

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
         "@types/chai": "^4.1.5",
         "@types/fs-extra": "^5.0.4",
         "@types/mocha": "^5.2.5",
-        "@types/node": "10.12.24",
+        "@types/node": "11.9.0",
         "@types/q": "^1.5.1",
         "@types/sinon": "7.0.5",
         "chai": "^4.2.0",
--- a/package.json
+++ b/package.json
@@ -357,6 +357,11 @@
                 "category": "Brightscript"
             },
             {
+                "command": "extension.brightscript.clearLogOutput",
+                "title": "Clear the brighscript log output",
+                "category": "Brightscript"
+            },
+            {
                 "command": "extension.brightscript.pressBackButton",
                 "title": "Press the Back button on the Roku remote",
                 "category": "Brightscript"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "brightscript",
     "displayName": "BrightScript Language",
-    "version": "1.8.2",
+    "version": "1.8.6",
     "publisher": "celsoaf",
     "description": "Language support for Roku's BrightScript language.",
     "author": {
@@ -229,6 +229,7 @@
                         "type": "brightscript",
                         "request": "launch",
                         "name": "BrightScript Debug: Launch",
+                        "stopOnEntry": true,
                         "host": "${promptForHost}",
                         "password": "${promptForPassword}",
                         "rootDir": "${workspaceFolder}"
@@ -245,8 +246,7 @@
                             "stopOnEntry": true,
                             "host": "${promptForHost}",
                             "password": "${promptForPassword}",
-                            "rootDir": "^\"\\${workspaceFolder}\"",
-                            "consoleOutput": "normal"
+                            "rootDir": "^\"\\${workspaceFolder}\""
                         }
                     }
                 ]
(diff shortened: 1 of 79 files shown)


georgejecook:
Hey @TwitchBronBron I've been thinking about how we could get this in line with your original proposition - perhaps add more formats that can handle the kind of output myself and others want (which is filename.method(lineNumber) )..

I didn't go that route, as I felt by doing that, I'd start codifying my direct needs into your extensions functionality, whereas this way is totally generic. I imagine there's more discussion to have on this feature - happy to do some tweaking

TwitchBronBron:
I believe there is a solution that accomplishes your your goals without needing to implement a custom pkg path format. While it doesn't provide the ultimate flexibility (changing the link text to literally whatever you want), I think it _does_ does accomplish your specific needs and probably most others' needs. 

Given `pkg:/components/BaseScene.brs:85`, we can open the BaseScene.brs file, go to line 85, and walk backwards line by line until we find its containing function. There's your function name. Then you have `BaseScene.DoSomething(85)` as a clickable link, without a need for a custom pkg format. If you can wait until the languageserver branch lands, it would be very easy to add a function like `getClosestNamedFunctionAtLine(lineNum)` which would make this trivial. Otherwise, we might just need to leverage your "find all function declarations" regex to identify the closest named function.

From our previous conversation on #75, we would add one additional format: `FilenameAndFunction` (open to changing the name). Obviously you don't need to implement all of these...you can just implement `Full` and `FilenameAndFunction` which provides the framework for myself or someone else to add the rest later with ease. 


| | |
|--|--|
|Full| `pkg:/components/KeyLogTester.brs(24:0) [INFO] retriving videos for category id 23`|
|FilenameAndFunction| `KeyLogTester.DoSomething(24:0) [INFO] retriving videos for category id 23`|
|Filename| `KeyLogtester.brs(24) [INFO] retriving videos for category id 23`|
|Short |`#1 [INFO] retriving videos for category id 23`|
|Hidden| `[INFO] retriving videos for category id 23`|

So my question to you: assuming my proposal is do-able, does it solve the way you want to see your logs?

georgejecook:
yeah - that does look good: how will it perform though? it's already a trifle slow.. would it be simple to put a cache around LSP?

TwitchBronBron:
The language server is really fast. However, its speed in general doesn't matter too much, because on extension startup it parses every file, and then stores the AST and all the function locations into memory for the lifetime of the app. Anytime a file changes, only that file is re-parsed. So, the `getClosestNamedFunctionAtLine` call should be almost instantaneous because already have all of that information available.

georgejecook:
Great. I have no time to do this yet - so I'll keep working with my half baked branch for a couple of weeks; then I'm happy to take it on if you like.

TwitchBronBron:
Sounds like a plan!

package.json
```diff
@@ -34,7 +34,7 @@
         "build-and-test": "npm run compile && npm run test",
         "test": "nyc mocha src/**/*.spec.ts --full-trace --bail",
         "coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov",
-        "postinstall": "node ./node_modules/vscode/bin/install",
+        "postinstall": "node .\node_modules/vscode/bin/install",
```
TwitchBronBron: This change doesn't look right. Those slashes mean the n is escaped, instead of being a directory separator.

package.json
```diff
@@ -61,7 +61,7 @@
         "@types/chai": "^4.1.5",
         "@types/fs-extra": "^5.0.4",
         "@types/mocha": "^5.2.5",
-        "@types/node": "11.9.1",
+        "@types\node": "11.9.1",
```
TwitchBronBron: Same issue as above. This is escaping the n character. This should be a forward slash.

package.json
```diff
@@ -182,7 +182,7 @@
                             },
                             "files": {
                                 "type": "array",
-                                "description": "An array of file paths, file globs, or {src:string;dest:string} objects that will be copied into the deployment package. This will override the defaults, so if specified, you must provide ALL files. See https://npmjs.com/roku-deploy for examples.",
+                                "description": "An array of file paths, file globs, or {src:string;dest:string} objects that will be copied into the deployment package. This will override the defaults, so if specified, you must provide ALL files. See https:/\npmjs.com/roku-deploy for examples.",
```
TwitchBronBron: another backslash in the npmjs url?

package.json
georgejecook: ugh - looks like a search and replace went astray. I'll fix it

package.json
georgejecook: addressed

package.json
georgejecook: addressed

TwitchBronBron:
@georgejecook I'm ready to merge this, but there are some conflicts due to the debug recovery PR. Can you get the conflicts resolved, and I'll accept the PR asap.

georgejecook:
Merged latest master and resoled conflicts.

TwitchBronBron:
fixes #128

Issue #68, opened 2019-01-14
CompletionItem Work


Romans-I-XVI:
Hi, so I am in the process of switching over from Sublime to VSCode. In sublime I had a bunch of snippets just to help me remember what interface methods are available in BrightScript. I began working on porting those snippets over and turning them in to CompletionItem data as part of this extension. You can see the work I've done so far here -> https://github.com/Romans-I-XVI/vscode-brightscript-language/tree/adding-completion-items

So what I have going so far is to detect if the line ends with a particular interface and give available methods. It's not optimal as I don't usually explicitly type out the interface I'm referencing, but I wasn't sure how else to know what suggested completion items to return. So for instance code would need to be typed as such.
```
a = CreateObject("roAppManager")
a.ifAppManager.
```

After typing that "." after ifAppManager you would be given a drop-down of methods available to that interface. 

You can see the general layout I have started in the fork I linked. My next steps are simply to work on implementing all available interface methods and the snippets & documentation for those methods. But before doing all that legwork I wanted to check in and see if there are already plans for how completion items will work, possibly in a more advanced way that what I have implemented as my implementation is pretty basic.

Thanks and best regards,
Austin

TwitchBronBron:
So I love this idea! as long as it doesn't produce false positives, I'm cool with your current approach.

In the medium-to-long-term, I hope to have a full language server written that would handle all of this auto completion stuff. However I would end up having to do all of the same legwork that you're doing anyway, so none of your work will be wasted.  Doing it your way would definitely provide benefit to this extension which would hold us over until the language server is completed (I just started writing it 3 days ago, so it will be a while...)

So go for it, and let us know if there's any way we can help!

Romans-I-XVI:
Oh ok awesome, I will continue work on it then and let you know if anything comes up.

PR #105, opened 2019-03-05
Adding Completion Items

Romans-I-XVI:
This implements the CompletionItem provider discussed in #68 

It adds snippets and documentation for all [BrightScript interface](https://sdkdocs.roku.com/display/sdkdoc/Interfaces) methods except for a few more obscure ones such as ifFunction as well as any interfaces/methods listed as being deprecated.

It also adds a basic language wide snippets file to provide easy completion of all available BrightScript interfaces.


Romans-I-XVI:
I'm not sure I understand why this check fails, I notice it will also fail with the constructor for vscode.SnippetString. Just let me know what I need to change and I will do so.

TwitchBronBron:
Thanks for this. I'll give this a look through in the next few days.

TwitchBronBron:
Hey. I did a quick scan through this (there's a lot!) and everything looks good to me. There was a merge conflict that I tried to resolve, but the build is still failing. Can you get it compiling again and push another change, and I'll merge it as soon as you do that?

Thanks for all of this work!

TwitchBronBron:
I forgot to mention, the issue is that I renamed the BrightScriptConfigurationProvider class to remove the word BrightScript. So I think you just need to remove that import from extension.ts

Romans-I-XVI:
Lol, yeah there's a lot... I had already done a about half of the work for a previous sublime text plugin though so that helped. I'll get master merged in and fix the conflicts.

src/BrightScriptCompletionItemProvider.ts
```diff
@@ -8,23 +8,136 @@ import {
...
+        ifTimespan: ifTimespanCompletionItems,
+        ifUrlTransfer: ifUrlTransferCompletionItems,
+        ifVideoPlayer: ifVideoPlayerCompletionItems,
+        ifXMLElement: ifXMLElementCompletionItems,
+        ifXMLList: ifXMLListCompletionItems
+    };
 
     public provideCompletionItems(document: TextDocument, position: Position, token: CancellationToken, context: vscode.CompletionContext): CompletionItem[] {
-        //TODO - do something useful here!
-        return this.BuiltinCompletionItems;
```
georgejecook: I'm deeply concerned by this - this looks to me like it won't _augment_ the existing contextual completions; but will *remove them* - that would mean we'd lose all the completion hints for our own vars and functions, which would greatly degrade the experience - I spent most my time calling my own method's and using my vars, not the interface ones, so I see nothing but pain if I lose the builtin completion items, which is why I didn't mess with this when I did the other definition and symbol work.

georgejecook:
Added a line comment - I was concerned that this would get rid of the default completions; but it seems like it doesn't when it returns undefined. saying that though - can we get some info on how this works ? I pulled the branch, and don't see how to get these completions.
 
My concerns are:

1. That these will be prioritized over default completions, which will stop me getting completions to my own code when there are collisions, instead of letting me choose the correct one
2. The casing : will this observe camelCase or enforce PascalCase
3. Many people use vscode for it's excellent default completions and definition support : I think fundamental changes to these should be thoroughly documented, tested and made available in beta, before being merged down.

TwitchBronBron:
@georgejecook take a look at the github issue associated with this PR (#68). For the completions in this effort, they only appear when you have typed the exact name of the interface after the variable name, like so: 
```
a = CreateObject("roAppManager")
a.ifAppManager.
```
In this case, the only valid completions that should be showing are the ones that @Romans-I-XVI now provides in this PR.

src/BrightScriptCompletionItemProvider.ts
TwitchBronBron: In the context of this PR, we do want to throw away completions in this case, because you have specifically typed the full name of the interface after your variable, (i.e. `yourVar.ifArray`) meaning there are no other valid completions except the ones that are provided in this PR.

Issue #74, opened 2019-01-16
variables view should populate vars on a breakpoint/runt time error


georgejecook:
we should get the output of a telnet vars command, in the vars panel in the top right, imo

chrisdp:
#117 up for review

chrisdp:
@georgejecook we can close this now yes?

PR #117, opened 2019-03-16
Feature/variable panel

chrisdp:


chrisdp:
#74 relates to this issue

georgejecook:
had a quick look at this, and it seems great - didn't cause me any problems : May I suggest a setting flag to turn it on or off though, in case it causes crashes/hangs in some scenarios?

chrisdp:
Sure. I’ll add that this morning.

TwitchBronBron:
Everything looks good here. I'm fine with the changes. @georgejecook can you give your final "ok" before I merge? I'm accepting the PR as is, with the decrease in unit test coverage, because the majority of the telnet and debugger stuff are VERY hard to test, and it's not worth the effort at this point.

georgejecook:
pulled down. works perfectly both enabled and disabled on 2 very large prod apps

PR #124, opened 2019-03-21
Close any leftover debugging threads before side loading

chrisdp:
This attempts to close any leftover debugging threads before side loading and when stopping the debug adapter.


TwitchBronBron:
Merging this now, will fix the tests in a later commit. 
I'm waiving the code coverage requirements because this is another hard one to test.

PR #129, opened 2019-03-26
Feature/skip onbreak false breakpoints

georgejecook:
guards against bogus breakpoints, by adding micro debugger awareness to adapter and log output manager.

bogus breakpoints result from `break in` style breakpoints.

Also adds additional settings `skipBogusBreakpoints` and `doNotIncludeStackTracesInLogOutput`, and wires them both into the log output and adapter so that they can be turned on/off

Tested both on and off for both flags, and observed no regressions in step debugging or breakpoint/resume and error stops

I can't guarantee this works in all cases : but I work with it on all the time now and the only issue I get is if I get impatient waiting for all the bogus breakpoints to pass, and hit another one.

### DIFF
--- a/src/LogOutputManager.ts
+++ b/src/LogOutputManager.ts
@@ -14,14 +14,18 @@ export class LogLine {
 }
 
 export class LogOutputManager {
+
     constructor(outputChannel, context) {
         this.collection = vscode.languages.createDiagnosticCollection('BrightScript');
         this.outputChannel = outputChannel;
+        this.config = vscode.workspace.getConfiguration('brightscript') || {};
         this.context = context;
         let subscriptions = context.subscriptions;
         this.includeRegex = null;
         this.logLevelRegex = null;
         this.excludeRegex = null;
+        this.debugStartRegex = new RegExp('BrightScript Micro Debugger\.', 'ig');
+        this.debugEndRegex = new RegExp('Brightscript Debugger>', 'ig');
 
         subscriptions.push(vscode.commands.registerCommand('extension.brightscript.markLogOutput', () => {
             this.markOutput();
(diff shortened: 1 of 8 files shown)


georgejecook:
@TwitchBronBron started this again. Hope this is okay - my log output pr is waiting on this to get merged

src/LogOutputManager.ts
```diff
@@ -145,11 +154,37 @@ export class LogOutputManager {
...
+                        console.log('ended MicroDebugger block');
+                        this.isInMicroDebugger = false;
+                        if (this.isNextBreakpointSkipped) {
+                            line = '\n**Was a bogus breakpoint** Skipping!\n';
+                        } else {
+                            line = null;
+                        }
+                    } else if (this.isInMicroDebugger) {
+                        this.currentMicroDebuggerText += line + '\n';
+                        if ((this.config.debug || {}).skipBogusBreakpoints && line.startsWith('Break in ')) {
```
TwitchBronBron: This line assumes the user set `skipBogusBreakpoints` in the user/workspace settings. However, a perfectly valid way of setting `skipBogusBreakpoints` is in the launch config (_without_ using settings at all)  meaning this code path won't always work with `skipBogusBreakpoints`. 

In [DebugConfigurationProvider around line 139](https://github.com/TwitchBronBron/vscode-brightscript-language/blob/master/src/DebugConfigurationProvider.ts#L102), we have a pattern of storing a value in workspaceState. I think this is a better approach. 

So, at the end of the debug config provider, run `this.context.workspaceState.update('skipBogusBreakpoints, config.skipBogusBreakpoints)`, and then in the code above, read that value ` await this.context.workspaceState.get('skipBogusBreakpoints')` instead of getting it from settings.

TwitchBronBron:
Looks pretty good. Thanks for your effort! I think we need to use a different word than "bogus". It doesn't feel quite right to me. Can you come up with a more descriptive name? Like `skipAsyncRunLoopBreakpoints` or `skipMicroDebuggerBreakpoints` or something? `bogus` could mean a lot of things, so if we could be a bit more specific, that would make the setting more official and also be more transparent about what it does at a glance.

georgejecook:
Updated with your requested changes.

TwitchBronBron:
Fixes #132

PR #130, opened 2019-03-27
Multiple source directory support 

NathanGillis:
Multiple source directory support 
- Depreciates debugrootDir string and replaces it with sourceDirs array to support multiple source directories. 
- Fires error when debugrootDir is used in the launch and prompts the user to update to sourceDirs.
- Removed all remaining breakpoints after looping through them to prevent crash on a breakpoint existing in the workspace but out of scope

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,12 +138,13 @@
                             },
                             "debugRootDir": {
                                 "type": "string",
-                                "description": "Depreciated, use sourceDirs, which is an array of source locations",
+                                "deprecationMessage":"Deprecated. Use sourceDirs instead",
+                                "description": "If you have a build system, rootDir will point to the build output folder, and this path should point to the actual source folder so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise debugger lines will be out of sync.",
                                 "default": "${workspaceFolder}"
                             },
                             "sourceDirs": {
                                 "type": "array",
-                                "description": "If you have a build system, sourceDirs will point to the build output folder, and this path should point to the actual source folder so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise debugger lines will be out of sync.",
+                                "description": "If you have a build system, this array will point to paths of actual source folders so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise, debugger lines will be out of sync.",
                                 "default": ["${workspaceFolder}"]
                             },
                             "outDir": {
--- a/package.json
+++ b/package.json
@@ -138,8 +138,13 @@
                             },
                             "debugRootDir": {
                                 "type": "string",
-                                "description": "If you have a build system, rootDir will point to the build output folder, and this path should point to the actual source folder so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise debugger lines will be out of sync.",
+                                "description": "Depreciated, use sourceDirs, which is an array of source locations",
                                 "default": "${workspaceFolder}"
+                            },
+                            "sourceDirs": {
+                                "type": "array",
+                                "description": "If you have a build system, sourceDirs will point to the build output folder, and this path should point to the actual source folder so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise debugger lines will be out of sync.",
+                                "default": ["${workspaceFolder}"]
                             },
                             "outDir": {
                                 "type": "string",
--- a/src/BrightScriptDebugSession.spec.ts
+++ b/src/BrightScriptDebugSession.spec.ts
@@ -366,7 +366,7 @@ describe('Debugger', () => {
 
         it('remaps to debug folder when specified', () => {
             (session as any).launchArgs = {
-                debugRootDir: path.normalize('/src'),
+                sourceDirs: [path.normalize('/src')],
                 rootDir: path.normalize('/dest')
             };
             args.breakpoints = [{ line: 1 }];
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -138,9 +138,14 @@
                             },
                             "debugRootDir": {
                                 "type": "string",
-                                "description": "If you have a build system, rootDir will point to the build output folder, and this path should point to the actual source folder so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise debugger lines will be out of sync.",
+                                "description": "Depreciated, use sourceDirs, which is an array of source locations",
```
TwitchBronBron: Leave the entire original message, but you can put the word "depreciated" at the front.

Also, as we discussed on Slack, add a `deprecationMessage` property and put this message in there. `Deprecated. Use sourceDirs instead`.

package.json
```diff
@@ -138,9 +138,14 @@
...
                                 "default": "${workspaceFolder}"
                             },
+                            "sourceDirs": {
+                                "type": "array",
+                                "description": "If you have a build system, sourceDirs will point to the build output folder, and this path should point to the actual source folder so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change line offsets between source files and built files, otherwise debugger lines will be out of sync.",
```
TwitchBronBron: This description doesn't sound correct. should read "**rootDir** will point to the build output folder..."

Also, you should add a brief explanation of how sourceDirs works when you have multiple array entries.

src/BrightScriptDebugSession.spec.ts
```diff
@@ -366,15 +366,15 @@ describe('Debugger', () => {
 
         it('remaps to debug folder when specified', () => {
             (session as any).launchArgs = {
-                debugRootDir: path.normalize('/src'),
+                sourceDirs: [path.normalize('/src')],
```
TwitchBronBron: Can you add another test that verifies `debugRootDir` is correctly converted to `sourceDirs`?

src/BrightScriptDebugSession.ts
```diff
@@ -550,11 +582,18 @@ export class BrightScriptDebugSession extends DebugSession {
                 //we found multiple files with the exact same path (unlikely)...nothing we can do about it.
             }
         }
-        //use debugRootDir if provided, or rootDir if not provided.
-        let rootDir = this.launchArgs.debugRootDir ? this.launchArgs.debugRootDir : this.launchArgs.rootDir;
+        let rootDir = this.launchArgs.sourceDirs ? this.launchArgs.sourceDirs : this.launchArgs.rootDir;
```
TwitchBronBron: `sourceDirs` is an array, but `rootDir` is a single string. So a `for of` loop on a string will loop through every character in the string. Please test to verify, but I think you need to wrap `this.launchArgs.rootDir` in brackets to put it into an array.

`let rootDir = this.launchArgs.sourceDirs ? this.launchArgs.sourceDirs : [this.launchArgs.rootDir];`

src/BrightScriptDebugSession.ts
```diff
@@ -826,7 +865,7 @@ interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
      * so that breakpoints can be set in the source files when debugging. In order for this to work, your build process cannot change
      * line offsets between source files and built files, otherwise debugger lines will be out of sync.
      */
-    debugRootDir: string;
+    sourceDirs: string[];
```
TwitchBronBron: We still need to have `debugRootDir` in this interface for now, until it actually gets removed, so be sure to add that back in, and add some type of @deprecated annotation or message on the comment.

src/DebugConfigurationProvider.ts
```diff
@@ -31,6 +31,12 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
     public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, config: BrightScriptDebugConfiguration, token?: CancellationToken): Promise<DebugConfiguration> {
         //make sure we have an object
         config = config ? config : {} as any;
+        
+        //Check for depreciated Items
+        if (config.debugRootDir){
```
TwitchBronBron: We still need to support debugRootDir for a while, so you should put `debugRootDir` INTO `sourceDirs` in this situation instead of throwing an error. Something like this:

```typescript
if(config.debugRootDir){
    if(config.sourceDirs){
        throw new Error("Cannot set both debugRootDir AND sourceDirs");
    } else{
        config.sourceDirs = [config.debugRootDir];
    }
}
```

src/DebugConfigurationProvider.ts
```diff
@@ -111,7 +117,7 @@ export interface BrightScriptDebugConfiguration extends DebugConfiguration {
     host: string;
     password: string;
     rootDir: string;
-    debugRootDir?: string;
+    sourceDirs?: string[];
```
TwitchBronBron: `debugRootDir` needs to remain in this interface for backwards compatibility.

src/LogDocumentLinkProvider.ts
```diff
@@ -20,10 +20,13 @@ export class LogDocumentLinkProvider implements vscode.DocumentLinkProvider {
         this.launchConfig = launchConfig;
         this.fileMaps = {};
 
-        let sourceRootDir = launchConfig.debugRootDir ? launchConfig.debugRootDir : launchConfig.rootDir;
-
+        let sourceRootDir = launchConfig.sourceDirs ? launchConfig.sourceDirs : launchConfig.rootDir;
```
TwitchBronBron: Same as above, verify, but I believe you need to wrap rootDir in array brackets so that `sourceRootDir` is always an array.

TwitchBronBron:
Great work! You kept it straightforward and clean. I have one request, which we already discussed on Slack. In order to maintain backwards compatibility, we need to still support `debugRootDir`. Once you have added that support back in and resolve the other small comments, I'll be happy to merge this.

TwitchBronBron:
Also, can you update the readme "Special Cases" "Debug source files with Custom build process" section explaining how the sourceDirs array works, and what problem it solves? You can either add a whole separate section, or just another paragraph at the bottom of that section with some additional information. People need to see examples, or they won't ever know this exists. :)

Issue #112, opened 2019-03-12
Support for conditional breakpoints


nsillik:
It would be great if we could add conditional breakpoints, like this:

![image](https://user-images.githubusercontent.com/141675/54224931-ca955900-44b7-11e9-8a61-c0c9faa3e0c5.png)

I was thinking it would be pretty easy to instead of just (`BrightScriptDebugSession.ts` line 605ish) doing:
```
lines[lineIndex] = `STOP\n${line} `;
```

We could instead do something (psuedocode):
```
if (breakpoint.isConditional()) {
  lines[lineIndex] = `if ${breakpoint.condition} then : STOP : end if\n${line} `;
} else {
  lines[lineIndex] = `STOP\n${line} `;
}
```

I ran into an issue when trying it out: `DebugProtocol.Breakpoint` doesn't have the conditions in it but `DebugProtocol.SourceBreakpoint` does. I haven't spent any time digging into it, so no idea how hard it'll be. I'll see if I can take a look soon though.

chrisdp:
@nsillik  Been testing this and I think I have it working in a very ruff form. I will  continue to explore this today.

chrisdp:
Fixed in #135

chrisdp:
@nsillik we can close this issue now correct? I added support for this a couple months ago unless you have been having issues with it?

PR #135, opened 2019-04-09
Support for conditional breakpoints

chrisdp:
PR is related to issue: #112


chrisdp:
Log point can take a format like the following: 
`"this is my message {dependency} {event}"` 
this is the equivalent of:
`print "this is my message"; {dependency}; " "; {event}; ""`

TwitchBronBron:
Looks great! Thanks for your work on this.
Fixes #112

Issue #137, opened 2019-04-10
AA keys not displayed properly when hovering


triwav:
If you have an AA with a key of either an `&` symbol or space it doesn't display properly. For the following:
```
"Food & Both":{...},
"Food&And":{...},
"Food Space":{...}
```

The output on hover is:
```
Both: <Component: roAssociativeArray>
Space: <Component: roAssociativeArray>
And: <Component: roAssociativeArray>
```

If you have multiple keys with the same last part that's actually displayed it will actually only return one of these results as well.

TwitchBronBron:
Fixed in 1.13.0

PR #141, opened 2019-04-11
High level objects not always parsed correctly

chrisdp:
This was related to using `exec` with a `g` flag. See the following from the docs:

If your regular expression uses the "g" flag, you can use the exec() method multiple times to find successive matches in the same string. When you do so, the search starts at the substring of str specified by the regular expression's lastIndex property (test() will also advance the lastIndex property).  Note that the lastIndex property will not be reset when searching a different string it will start its search at its existing lastIndex .

### DIFF
--- a/src/RokuAdapter.ts
+++ b/src/RokuAdapter.ts
@@ -639,9 +639,29 @@ export class RokuAdapter {
         });
     }
 
-    private expressionRegex = /([\s|\S]+?)(?:\r|\r\n)+brightscript debugger>/i;
-    private isObjectCheckRegex = /<.*:\s*(\w+\s*\:*\s*\w*)>/gi;
-    private getFirstWordRegex = /^([\w.\-=]*)\s/;
+    /**
+     *
+     * @param value
+     */
+    private getExpressionDetails(value: string) {
+        return new RegExp(/([\s|\S]+?)(?:\r|\r\n)+brightscript debugger>/i).exec(value);
+    }
+
+    /**
+     * Runs a regex to check if the target is an object and get the type if it is
+     * @param value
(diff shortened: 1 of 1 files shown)


chrisdp:
Has fix for issue #137

src/RokuAdapter.ts
```diff
@@ -639,9 +639,29 @@ export class RokuAdapter {
...
+        return new RegExp(/<.*:\s*(\w+\s*\:*\s*[\w\.]*)>/gi).exec(value);
+    }
+
+    /**
+     * Runs a regex to get the first work of a line
+     * @param value
+     */
+    private getFirstWord(value: string) {
+        return new RegExp(/^([\w.\-=]*)\s/).exec(value);
+    }
```
chrisdp: Changed the above into functions to make it a bit harder to accidentally break a regex that is used in many places by simply adding a g flag.

src/RokuAdapter.ts
```diff
@@ -792,7 +812,7 @@ export class RokuAdapter {
                     return children;
                 }
                 let match;
-                match = /(\w+):(.+)/i.exec(line);
+                match = /(\S+[^:]):(.+)/i.exec(line);
```
chrisdp: This adds support for keys like this:
```
"Food & Both":{...},
"Food&And":{...},
"Food Space":{...}
```

TwitchBronBron: What about keys with colons in them? 

`"my:key"`

chrisdp: While possible I'm not sure how we could detect that because of the way the Roku prints key to the console.

Issue #148, opened 2019-05-13
Can't create pure screensaver


John-Whitehead:
A 'pure' Roku screensaver uses RunScreenSaver() as its entry point.  There are no main() or RunUserInterface() functions in a pure screensaver.  A pure screensaver is not listed as a channel, only on the Screensavers page.

I created a pure screensaver using the extension and when I go to run it I see an error - "Unable to find an entry point. Please make sure that you have a RunUserInterface or Main sub/function declared in your BrightScript project."

Please update the extension to support pure screensavers.

chrisdp:
@TwitchBronBron This should now be supported with the merge of PR #149 Can this issue be closed?

TwitchBronBron:
@chrisdp yeah, it should be closed. Thanks.

PR #149, opened 2019-05-13
Run screen saver entrypoint check

airving10:
Related to issue #148


TwitchBronBron:
Looks pretty good. There are a few other locations that set the `stopOnEntry` default value. I think in the package.json. Can you update those as well?

PR #150, opened 2019-05-14
adds enableAutoTerminate launchArg, which will close the debug session, if set to true

georgejecook:


TwitchBronBron:
@georgejecook PR looks great. Just a few small changes:
 - The default setting should be to stop debugging at the end of a single program run. Almost every debugger I've used in the past works this way (NodeJS, C#, VB, Python, Golang, Java, within netbeans, vscode, visual studio, eclipse, etc...). For developers who want to keep debugging when relaunching the app with the remote...it's easy enough to change this debug config option.
 - Change the name of `enableAutoTerminate` to something like `stopDebuggerOnAppExit`,  `endDebugSessionOnAppExit`, `stopDebuggingOnExit`, `keepDebuggingAfterAppExit`. Open to sugestions here.

georgejecook:
k, sounds good - I was specifically asked by Chris to set it to default off; but I prefer it with default on, too.

I'll get these changes in at lunch/or tonight.

chrisdp:
Was just thinking that users who have been using the ext but haven't been following the development could be confused by the change. Not a big deal ether way though for me. I'm ok with it being changed to on by default.

georgejecook:
@TwitchBronBron blimey! that took some time.. sorry for taking so long to get this done.. got another on the back of this too.

PR #151, opened 2019-05-15
Feature/lookup node children

georgejecook:
adds flag to enable resolving children of nodes variables - this is on top of #150 , coz I didn't have time to go and split this out - so we can review/ merge this after #150


package.json
```diff
@@ -264,7 +269,9 @@
                         "host": "${promptForHost}",
                         "password": "${promptForPassword}",
                         "rootDir": "${workspaceFolder}",
-                        "enableDebuggerAutoRecovery": false
+                        "enableDebuggerAutoRecovery": false,
+                        "enableAutoTerminate": false,
```
TwitchBronBron: @georgejecook I think this is the OLD name for `stopDebuggerOnAppExit`

package.json
```diff
@@ -279,7 +286,9 @@
                             "host": "${promptForHost}",
                             "password": "${promptForPassword}",
                             "rootDir": "^\"\\${workspaceFolder}\"",
-                            "enableDebuggerAutoRecovery": false
+                            "enableDebuggerAutoRecovery": false,
+                            "enableAutoTerminate": false,
```
TwitchBronBron: Change this to 'stopDebuggerOnAppExit' like the comment above?

src/RokuAdapter.ts
```diff
@@ -156,9 +153,10 @@ export class RokuAdapter {
     /**
      * Connect to the telnet session. This should be called before the channel is launched.
      */
-    public async connect(enableDebuggerAutoRecovery: boolean = false) {
+    public async connect(enableDebuggerAutoRecovery: boolean = false, enableLookupVariableNodeChildren: boolean = false) {
```
TwitchBronBron: @georgejecook Is there a reason you add these parameters on `connect` instead of in the constructor? If not, could you move both of them into the constructor so it's consistent? Feel free to convert the constructor into a single "options" parameter.

package.json
georgejecook: good catch - merge issue. thx

package.json
georgejecook: done

src/RokuAdapter.ts
georgejecook: I think I got confused when I was doing the original implementation for enableDebuggerAutoRecovery, and thought it had to be set in connect - just verified both flags work fine in the constructor.

georgejecook:
addressed your pr feedback @TwitchBronBron

PR #154, opened 2019-05-24
Add region code folding

nishtahir:
This should result in a fold in the editor.
```
'#region
sub doAThing()
...
end sub
'#endregion
```


TwitchBronBron:
Well that was a nice easy fix. Great work!

PR #156, opened 2019-05-31
Component library support

chrisdp:
This is very early support for component libraries. I am interested in feedback as this is the first time I have worked on static file hosting and file manipulation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,17 +169,69 @@
                             },
                             "componentLibraries": {
                                 "type": "array",
-                                "description": "An array of file path sets. One for each component library. Each index is an array of file paths, file globs, or {src:string;dest:string} objects that will be copied into the hosted component library. This will override the defaults, so if specified, you must provide ALL files. See https://npmjs.com/roku-deploy for examples. You must specify a componentLibrariesOutDir to use this.",
-                                "default": [{
-                                    "rootDir": "${workspaceFolder}/LibDir/",
-                                    "outFile": "LibName.zip",
-                                    "files": [
-                                        "manifest",
-                                        "source/**/*.*",
-                                        "components/**/*.*",
-                                        "images/**/*.*"
-                                    ]
-                                }]
+                                "description": "An array of file path sets. One for each component library. Each index is an array of file paths, file globs. This will override the defaults, so if specified, you must provide ALL files. See https://npmjs.com/roku-deploy for examples. You must specify a componentLibrariesOutDir to use this.",
+                                "default": [
+                                    {
+                                        "rootDir": "${workspaceFolder}/LibDir/",
+                                        "outFile": "LibName.zip",
+                                        "files": [
+                                            "manifest",
+                                            "source/**/*.*",
+                                            "components/**/*.*",
+                                            "images/**/*.*"
+                                        ]
+                                    }
+                                ],
+                                "items": {
+                                    "type": "object",
+                                    "description": "Component Library Schema.",
+                                    "default": {
+                                        "rootDir": "${workspaceFolder}/LibDir/",
+                                        "outFile": "LibName.zip",
+                                        "files": [
+                                            "manifest",
+                                            "source/**/*.*",
+                                            "components/**/*.*",
+                                            "images/**/*.*"
+                                        ]
+                                    },
+                                    "required": [
+                                        "rootDir",
+                                        "outFile",
+                                        "files"
+                                    ],
+                                    "properties": {
+                                        "rootDir": {
+                                            "type": "string",
+                                            "description": "Root directory of source files for this component library.",
+                                            "examples": [
+                                                "${workspaceFolder}/LibDir/"
+                                            ]
+                                        },
+                                        "outFile": {
+                                            "type": "string",
+                                            "description": "Output file name. Used as the static file name for the zip.",
+                                            "default": "",
+                                            "examples": [
+                                                "LibName.zip"
+                                            ],
+                                            "pattern": "^(.*)(\\.zip)$"
+                                        },
+                                        "files": {
+                                            "type": "array",
+                                            "description": "A file path or file glob that should be copied to the deployment package.",
+                                            "default": [
+                                                "manifest",
+                                                "source/**/*.*",
+                                                "components/**/*.*",
+                                                "assets/**/*.*"
+                                            ],
+                                            "items": {
+                                                "type": "string"
+                                            }
+                                        }
+                                    }
+                                }
                             },
                             "outDir": {
                                 "type": "string",
--- a/package.json
+++ b/package.json
@@ -170,7 +170,16 @@
                             "componentLibraries": {
                                 "type": "array",
                                 "description": "An array of file path sets. One for each component library. Each index is an array of file paths, file globs, or {src:string;dest:string} objects that will be copied into the hosted component library. This will override the defaults, so if specified, you must provide ALL files. See https://npmjs.com/roku-deploy for examples. You must specify a componentLibrariesOutDir to use this.",
-                                "default": []
+                                "default": [{
+                                    "rootDir": "${workspaceFolder}/LibDir/",
+                                    "outFile": "LibName.zip",
+                                    "files": [
+                                        "manifest",
+                                        "source/**/*.*",
+                                        "components/**/*.*",
+                                        "images/**/*.*"
+                                    ]
+                                }]
                             },
                             "outDir": {
                                 "type": "string",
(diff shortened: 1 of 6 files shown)


chrisdp:
[CompLibTestChannel.zip](https://github.com/TwitchBronBron/vscode-brightscript-language/files/3244977/CompLibTestChannel.zip)
This is a sample channel you can use to test this. It is currently set up so that you open VS Code with the compWorkspace.code-workspace file to open it as a work space of two nested work spaces but you also could just have the launch config one level higher in the folder structure. 

I have it done this way to simulate my work environment where we have a demo channel and the comp lib checked out in parallel with there own .vscode fold with launch configs and such.

NathanGillis:
I tested this PR out and it works!!!! Nice Job @chrisdp. Breakpoints in component library NICE.

package.json
```diff
@@ -156,6 +157,30 @@
                                     "${workspaceFolder}"
                                 ]
                             },
+                            "componentLibraryPort": {
```
TwitchBronBron: Let's make this plural like the other ones. `componentLibrariesPort`.

src/BrightScriptDebugSession.ts
```diff
@@ -287,6 +306,157 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                let stagingFolder = await this.rokuDeploy.prepublishToStaging(componentLibrary);
+                let paths = glob.sync(path.join(stagingFolder, '**/*'));
+                let pathDetails: object = {};
+
+                // Add breakpoint lines to the staging files and before publishing
+                this.sendDebugLogLine('Adding stop statements for active breakpoints in Component Libraries');
+                this.convertBreakpointPaths(componentLibrary.rootDir, componentLibrary.rootDir);
+                await this.addBreakpointStatements(stagingFolder, componentLibrary.rootDir);
+
+                for (let filePath of paths) {
```
TwitchBronBron: I know it makes this code a little more complicated, but we should consider running each iteration in parallel (not actual parallel, but the node style). So something like this: 

```typescript
await Promise.all(paths.map(async (filePath)=>{
    ....the for loop body
{));
```

This will allow things to complete a bit faster because we are doing work as soon as we are free, instead of iterating through each path in order.

src/BrightScriptDebugSession.ts
```diff
@@ -287,6 +306,157 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                                files: [
+                                    path.join(stagingFolder, '**/*.xml'),
+                                    path.join(stagingFolder, '**/*.brs')
+                                ],
+                                from: (file) => new RegExp(parsedPath.base, 'gi'),
+                                to: newFileName
+                            });
+
+                            // Rename the brs files to include the postfix name spacing tag
+                            fs.rename(filePath, path.join(stagingFolder, relativePath), function(err) {
```
TwitchBronBron: These callbacks are not awaited, so technically we could encounter race conditions where the files haven't been totally renamed yet. 

In addition to the previous request to make this for loop async/parallel, you should consider using a promise-based rename function, and awaiting it. Check out the `move` function from the [fs-extra](https://github.com/jprichardson/node-fs-extra/blob/master/docs/move.md) package. You can do something like this:

```typescript
await fsExtra.move(source, dest);
```

src/BrightScriptDebugSession.ts
```diff
@@ -287,6 +306,157 @@ export class BrightScriptDebugSession extends DebugSession {
         }
     }
 
+    private componentLibraryPostfix: string = '__lib';
+
+    protected async prepareAndHostComponentLibraries(componentLibraries, componentLibrariesOutDir: string, port: number) {
+        if (componentLibraries && componentLibrariesOutDir) {
+            this.componentLibrariesOutDir = componentLibrariesOutDir;
+            this.componentLibrariesStagingDirPaths = [];
+            let replace = require('replace-in-file');
```
TwitchBronBron: Move this to the top of the file with the other imports, and use the typescript import syntax. 
```typescript
import * as replaceInFile from 'replace-in-file';
```

src/BrightScriptDebugSession.ts
```diff
@@ -287,6 +306,157 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                    if (parsedPath.ext) {
+                        let originalRelativePath = relativePath;
+
+                        if (parsedPath.ext === '.brs') {
+                            // Create the new file name to be used
+                            let newFileName: string = `${parsedPath.name}${this.componentLibraryPostfix}${libraryNumber}${parsedPath.ext}`;
+                            relativePath = path.join(parsedPath.dir, newFileName);
+
+                            // Update all the file name references in the library to the new file names
+                            replace.sync({
```
TwitchBronBron: Use an async promise-returning version of this function

```typescript
await replace(...yourStuff);
```

src/BrightScriptDebugSession.ts
```diff
@@ -287,6 +306,157 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                '.mp3': 'audio/mpeg',
+                '.svg': 'image/svg+xml',
+                '.pdf': 'application/pdf',
+                '.doc': 'application/msword',
+                '.eot': 'appliaction/vnd.ms-fontobject',
+                '.ttf': 'aplication/font-sfnt',
+                '.zip': 'application/zip'
+            };
+
+            http.createServer((req, res) => {
```
TwitchBronBron: Can you break the http server stuff out into its own class?

src/BrightScriptDebugSession.ts
```diff
@@ -287,6 +306,157 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                            // based on the URL path, extract the file extension. e.g. .js, .doc, ...
+                            const ext = path.parse(pathname).ext;
+                            // if the file is found, set Content-type and send data
+                            res.setHeader('Content-type', mimeType[ext] || 'text/plain' );
+                            res.end(data);
+                        }
+                    });
+                });
+
+            }).listen(port);
```
TwitchBronBron: What happens if we restart the debug session (instead of stop and start). Does the http server get completely discarded and completely recreated? I'm just concerned that a restart will error out because the http server and port are already occupied.

chrisdp:
In addressing feedback I discovered a problem with breakpoints that I will need to address. It seems like the breakpoints are no longer being added into the source channel. This is related to the logic in the `deleteBreakpointsForClientPath` function. I will look into this after addressing the more simple feedback.

src/BrightScriptDebugSession.ts
chrisdp: Moving to this type of import seems to be giving me a problem. 
image

chrisdp: Not sure really why that is though.

src/BrightScriptDebugSession.ts
chrisdp: When trying to move to the async version I ran into another problem. I attempted to understand the issue but sadly I wasn't able to figure it out. This is following the import change in the comment above.
image

src/BrightScriptDebugSession.ts
chrisdp: Have not had any problems with the server and or port being occupied after debugger stops or restarts.

chrisdp:
I managed to fix the breakpoint issue and other feedback. We are good for another round of review 😃

package.json
```diff
@@ -156,6 +157,30 @@
...
+                                "description": "Port to access component libraries.",
+                                "default": 8080
+                            },
+                            "componentLibrariesOutDir": {
+                                "type": "string",
+                                "description": "Output folder the component libraries will be hosted in.",
+                                "default": "${workspaceFolder}/libs"
+                            },
+                            "componentLibraries": {
+                                "type": "array",
```
TwitchBronBron: Sorry for not catching this earlier. Using type "array" isn't good enough here, because the user won't get any intellisense for `rootDir`, `outFile`, `files`, etc. You should copy the `items` section for `files` down below into this. Then test and verify that in vscode, it provides intellisense for each of those items.

TwitchBronBron:
@chrisdp This looks great! Fantastic work. I just have the one comment, (adding json schema info for the componentLibraryFiles" section, and then we're good to go.

TwitchBronBron:
@chrisdp Oh, another thing. Can you please update the readme with your new debug config settings, as well as a section explaining how this stuff works? If the user doesn't know about it...then it doesn't exist. :D

chrisdp:
@TwitchBronBron I will update the README on usage first thing tomorrow. I have already pushed up the schema you requested so feel free to take a look at that.

PR #161, opened 2019-06-14
Simple service discovery protocol support

chrisdp:


chrisdp:
Added a small fix where the quick picker would be shown when there is not devices in the active device cache rather than going straight to the device IP input box. I am not expecting any more changes to this PR other than potential feedback changes.

TwitchBronBron:
@chrisdp This looks great. A few small things:
1. we are waiting to hear from the community about whether the popups should be enabled or disabled by default.
2. Can you update the readme with the new user settings and explain the functionality in general.

README.md
```diff
@@ -196,6 +196,8 @@ This extension contributes the following settings:
 * `brightscript.output.clearOnLaunch`: If set to true, will clear the brigthscript log when launching
 * `brightscript.output.clearConsoleOnChannelStart`: If set to true, will clear the  brigthscript log after connecting to the Roku channel after launching
 * `brightscript.output.hyperlinkFormat`: specifies the display format for log output `pkg` link
+* `brightscript.deviceDiscovery.showInfoMessages`: If set to true, a info toast will be shown when a Roku device has been found on the network.
+* `brightscript.deviceDiscovery.enabled`: If set to true, the extension will automatically watch and scan the network for online Roku devices. This can be pared with the `${promptForHost}` option in the launch config to display a list of online Rokus removing the need to constantly change the host IP in your config files.
```
chrisdp: @TwitchBronBron README was updated.

PR #169, opened 2019-07-04
Rendezvous tracking

chrisdp:


chrisdp:
@TwitchBronBron Any idea why travis would fail on the import of the node module I added? 
```
import * as arraySort from 'array-sort';
                                        ^
TypeError: Class extends value undefined is not a constructor or null
```

Not really sure why that is.

TwitchBronBron:
@chrisdp It might be related to the version of node you are using locally versus the version of node running travisci (although doesn't seem like that's the case) my travis build is using  "11.4.0". Perhaps try using nvm to install exactly that version locally and see if it causes the same issues?

chrisdp:
Just updated with master.

airving10:
A clear feature would be very helpful when testing specific aspects of a channel.

chrisdp:
Thanks for that. That would be very easy to add as and extension command. I will look into adding that tonight or tomorrow.

src/BrightScriptDebugSession.ts
```diff
@@ -248,7 +279,7 @@ export class BrightScriptDebugSession extends DebugSession {
                 }
 
                 this.sendEvent(new CompileFailureEvent(compileErrors));
-                //TODO - shot gracefull
+                //TODO - shot graceful
```
TwitchBronBron: Not that this is a big deal, but i think the previous text was more grammatically correct. lol How about we just change it to 

```//stop the roku adapter and exit the channel```

File: src/RendezvousTracker.ts [hunk 2]
```diff
@@ -0,0 +1,268 @@
...
+     * @returns The debug output after parsing
+     */
+    public processLogLine(logLine: string): string {
+        let dataChanged = false;
+        let lines = logLine.split('\n');
+        let normalOutput = '';
+
+        lines.map((line) => {
+            let match;
+            // see the following for an explanation for this regex: https://regex101.com/r/In0t7d/5
```
TwitchBronBron: Nice! i love the saved regex101 link here. Very helpful.

src/RendezvousTracker.ts
```diff
@@ -0,0 +1,268 @@
...
+                    // detected the completion of a rendezvous event
+                    dataChanged = true;
+                    let blockInfo = this.rendezvousBlocks[id];
+                    let clientLineNumber: string = this.clientPathsMap[blockInfo.fileName].clientLines[blockInfo.lineNumber].toString();
+
+                    if (this.rendezvousHistory[blockInfo.fileName]) {
+                        // file is in history
+                        if (this.rendezvousHistory[blockInfo.fileName][clientLineNumber]) {
+                            // line is in history, just update it
+                            (this.rendezvousHistory[blockInfo.fileName][clientLineNumber] as RendezvousLineInfo).totalTime += this.getTime(duration);
```
TwitchBronBron: Please find a way to not need to type cast this object. I think we can make the interfaces more clear instead.

src/RendezvousTracker.ts
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * Used to see if a field on a object in RendezvousHistory object is a reserved field
+ * @param fieldName name of the field to check
+ */
+export function isRendezvousDetailsField(fieldName: string): boolean {
+    return (fieldName === 'type' || fieldName === 'hitCount' || fieldName === 'totalTime' || fieldName === 'zeroCostHitCount');
+}
+
+export interface RendezvousHistory {
```
TwitchBronBron: Delete this interface and instead, just use `{[key: string]: RendezvousFileInfo}` where it's defined above.

src/RendezvousTracker.ts
```diff
@@ -0,0 +1,268 @@
...
+    [key: string]: RendezvousFileInfo;
+}
+
+interface RendezvousFileInfo {
+    [key: string]: RendezvousLineInfo | ElementType | number;
```
TwitchBronBron: Don't make this object indexable like this. From what I can tell, the indexable item is the line information, so instead, add a property to this interface (and the actual objects where created) called `lines` or something. 

Also, since you are the one constructing this object, each property should be a single object type (instead of the three that it is right now). If there are multiple things, they should each get their own property on this object.

src/RendezvousTracker.ts
```diff
@@ -0,0 +1,268 @@
...
+        return duration ? parseFloat(duration) : 0.000;
+    }
+
+    /**
+     * Replaces the last occurrence of a string in a string
+     * @param base String to search within
+     * @param replace String to replace
+     * @param replacement What to replace the final occurrence with
+     */
+    private replaceLastStringOccurrence(base: string, replace: string, replacement: string): string {
```
TwitchBronBron: Here's a node module for this functionality, so try using that isntead: https://www.npmjs.com/package/replace-last

If that doesn't work, then at least move this function into `util.ts`.

package.json
```diff
@@ -110,6 +111,30 @@
...
+                    "name": "Rendezvous",
+                    "when": "debugType == 'brightscript'",
+                    "group": "Rendezvous"
+                }
+            ]
+        },
+        "menus": {
+            "view/title": [
+                {
+                    "command": "extension.brightscript.toggleSortingMethod",
```
TwitchBronBron: let's call it `extension.brightscript.toggleRendezvousSortingMethod`, or `extension.brightscript.rendezvous.toggleSortingMethod

src/RendezvousTracker.ts
chrisdp: Figured there was no need for a import if we could simply write the one function our selfs. I can move to utils for sure.

package.json
```diff
@@ -110,6 +111,30 @@
...
+                    "when": "debugType == 'brightscript'",
+                    "group": "Rendezvous"
+                },
+                {
+                    "command": "extension.brightscript.toggleSortingAscDesc",
```
TwitchBronBron: Let's call this `extension.brightscript.toggleRendezvousSortDirection` or `extension.brightscript.rendezvous.toggleSortDirection`

src/RendezvousTracker.ts
chrisdp: I wanted to keep the overall structure as close as possible to what the view was going to use to keep it as simple as possible and as one to one.

src/RendezvousViewProvider.ts
```diff
@@ -0,0 +1,322 @@
...
+        return (!this.sortAscending) ? sortResult : (sortResult === 1) ? -1 : (sortResult === -1) ? 1 : sortResult;
+    }
+
+    /**
+     * With return the differences in two objects
+     * @param obj1 base target
+     * @param obj2 comparison target
+     * @param exclude fields to exclude in the comparison
+     */
+    private objectDiff(obj1: object, obj2: object, exclude?: string[]) {
```
TwitchBronBron: Can you move this into `util.ts`?

src/RokuAdapter.ts
```diff
@@ -1010,6 +1022,21 @@ export class RokuAdapter {
             } while (commandsExecuted < 10);
         }
     }
+
+    /**
+     * Passes the debug functions used to locate the client files and lines to the RendezvousTracker
+     */
+    public setRendezvousDebuggerFileConversionFunctions(convertDebuggerLineToClientLine, convertDebuggerPathToClient) {
```
TwitchBronBron: Can you please define the types for these parameters. ` (debuggerPath: string, lineNumber: number)=> number` and `(debuggerPath: string) => string` (I think).

package.json
chrisdp: Ok, I think I am going to add a clear command as well under the command `extension.brightscript.rendezvous.clearHistory` as per @airving10 comment. This would be a very simple command to add and I think it is very worth adding.

src/RendezvousTracker.ts
chrisdp: Also I did this so that we wouldn't have to search arrays for existing data every-time since we should just do a key lookup to quickly check. Storing the data in arrays makes the process of updating the view much harder.

TwitchBronBron: I wasn't saying to use arrays instead of objects, i was just saying that when the root object is indexable, it's not clear what that index is actually holding, so I would prefer it to exist in a named property instead indexing the root object. 

If this is the exact structure required for the vscode tree view, then I'm ok with leaving it as is...but at least add some comments on this to explain what exactly it's holding and why it is done this way.

src/RendezvousViewProvider.ts
chrisdp: I can and will.

src/RendezvousTracker.ts
TwitchBronBron: I don't mind using additional modules because they reduce the amount of code we need to maintain ourselves. The one I linked to above has zero dependencies.

src/RendezvousTracker.ts
chrisdp: Ah, ok. I see what you'er saying. I think I will move it into an `occurrences` field as currently we index by file and then also internally index by line in the file. This way I can keep the location to iterate over in the view the same key name.

src/RendezvousTracker.ts
chrisdp: Ah, good point. Ok I will move over to the module.

src/RendezvousTracker.ts
chrisdp: This interface exists as it is used in the RendezvousViewProvider.

src/BrightScriptDebugSession.ts
chrisdp: Done

src/RendezvousTracker.ts
chrisdp: Done

src/RendezvousTracker.ts
chrisdp: Updated to make this interface more useful down the line.

src/RendezvousTracker.ts
chrisdp: Done

src/RendezvousTracker.ts
chrisdp: Done

package.json
chrisdp: Done

package.json
chrisdp: Done

src/RokuAdapter.ts
chrisdp: Done and addressed in the RendezvousTracker as well.

src/RendezvousViewProvider.ts
chrisdp: Done

chrisdp:
This PR is now dependent on #172.
Once that is merged this one should be good to merge as well.

src/RendezvousTracker.ts
```diff
@@ -0,0 +1,290 @@
...
+interface RendezvousClientPathMap {
+    [key: string]: RendezvousClientFile;
+}
+
+interface RendezvousClientFile {
+    [key: string]: string | RendezvousClientLineMap;
+    clientPath: string;
+}
+interface RendezvousClientLineMap {
+    [key: string]: number;
```
TwitchBronBron: This `RendezvousClientLineMap` interface is unnecessary. Just inline the type at the locations it's used and remove the interface.

src/RendezvousTracker.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+type ElementType = 'historyInfo' | 'fileInfo' | 'lineInfo';
+
+interface RendezvousClientPathMap {
+    [key: string]: RendezvousClientFile;
+}
+
+interface RendezvousClientFile {
+    [key: string]: string | RendezvousClientLineMap;
```
TwitchBronBron: This is still a bit muddy as far as typing is concerned. Do yo ureally need to make this a map that holds two separate things? Is there a way to break this out so that you have a property for whatever holds the [key:string]:string, and separate property for the [key:string]:number? I'm open to hearing a justifcation for this, but I try very hard to restrict properties to represent a single type. And it's not like `string | number`, this is `dictionary of string` OR `dictionary of dictionary of number`, which makes it a bit more confusing.

src/RendezvousViewProvider.ts
```diff
@@ -0,0 +1,272 @@
...
+
+export class RendezvousViewProvider implements vscode.TreeDataProvider<vscode.TreeItem> {
+
+    constructor(context: vscode.ExtensionContext) {
+        this.enableSmartSorting();
+
+        // #region Register sorting commands
+        let subscriptions = context.subscriptions;
+        subscriptions.push(vscode.commands.registerCommand('extension.brightscript.rendezvous.toggleSortMethod', () => {
+            if (!this.enableSmartSorting()) {
```
TwitchBronBron: This is a little confusing, because the actions are also returning state. Can you try to separate the `enable/disable` functionality from the `is it enabled / disabled` for this condition? Something like:
```
this.enableSmartSorting();
if (this.smartSortingIsEnabled === false) {
    this.disableSmartSorting();
}
```

Also, can you add a comment here explaining why calling `enableSmartSorting` would do anything other than enable smart sorting, and why you even need to check? Perhaps the naming of the functions should be updated, like `tryEnableSmartSorting`, or break out the functionality so that enable actually always enables, and you do conditional checks around it instead.

src/RendezvousViewProvider.ts
```diff
@@ -0,0 +1,272 @@
...
+                    return new RendezvousFileTreeItem(key, vscode.TreeItemCollapsibleState.Collapsed, null, this.rendezvousHistory.occurrences[key]);
+                }), this.activeFilter);
+            } else {
+                return [];
+            }
+        } else {
+            // VS code is asking for the children of the supplied tree item
+            let treeElement = this.getTreeElementHistoryData(element);
+
+            let result;
```
TwitchBronBron: Can you give the result a type here? It defaults to `any` in this situation.

src/RokuAdapter.ts
```diff
@@ -1144,7 +1196,7 @@ export class RequestPipeline {
     }
 
     /**
-     * Internall request processing function
+     * Internally request processing function
```
TwitchBronBron: I think the word should be `Internal`. Spell checker's getting a little aggressive. :P

src/RendezvousViewProvider.ts
chrisdp: Changing this to a `toggleSmartSorting` function that will simply swap between the two.

src/RendezvousTracker.ts
chrisdp: This interface has been combined into the RendezvousClientPathMap.

src/RendezvousTracker.ts
chrisdp: This interface definition has been improved for readability.

chrisdp:
Feedback have been addressed and is ready for another round of review when your ready @TwitchBronBron

PR #172, opened 2019-07-17
First pass at correcting new lines in the output and some logs not ma…

chrisdp:
…king it to the output and debug consoles.

### DIFF
--- a/src/RokuAdapter.ts
+++ b/src/RokuAdapter.ts
@@ -1065,7 +1065,6 @@ export class RequestPipeline {
     }
 
     private requests: RequestPipelineRequest[] = [];
-    private waitForPromptRequests: RequestPipelineRequest[] = [];
     private debuggerLineRegex: RegExp;
     private isAtDebuggerPrompt: boolean = false;
 
@@ -1074,7 +1073,7 @@ export class RequestPipeline {
     }
 
     private get hasRequests() {
-        return this.waitForPromptRequests.length > 0;
+        return this.requests.length > 0;
     }
 
     private currentRequest: RequestPipelineRequest = undefined;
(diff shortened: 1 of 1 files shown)


src/RokuAdapter.ts
```diff
@@ -1128,43 +1165,52 @@ export class RequestPipeline {
...
                 },
-                waitForPrompt: waitForPrompt
-            });
-            //start processing (safe to call multiple times)
-            this.process();
+                waitForPrompt: waitForPrompt,
+            };
+
+            if (!waitForPrompt) {
+                if (!this.isProcessing) {
```
TwitchBronBron: @chrisdp  There's no `else` condition for this...so there's a chance that `waitForPrompt === false && this.isProcessing === true`...that won't be handled by anything. That may cause issues.

src/RokuAdapter.ts
```diff
@@ -1128,43 +1165,52 @@ export class RequestPipeline {
...
+                    //fire and forget the command
+                    request.executeCommand();
+                    //the command doesn't care about the output, resolve it immediately
+                    request.onComplete(undefined);
+                }
+            } else {
+                this.requests.push(request);
+                if (this.isAtDebuggerPrompt) {
```
TwitchBronBron: Same issue as above. What happens when `waitForPrompt === true && this.isAtDebuggerPrompt === false`

src/RokuAdapter.ts
chrisdp: That is fine as this is what used to happen before due to the old check in the process function:
```
        if (this.isProcessing || this.requests.length === 0) {
            return;
        }
```

chrisdp: But I guess it would just get processed later.
I think It might actutatly be find the way I have it to keep adding them to the que as only the wait for debugger ones will trigger processing with this change.

chrisdp:
Pushed updated related to our slack conversation. Should be good for review again.

PR #179, opened 2019-07-22
Better errors during publishing process

chrisdp:


src/extension.ts
```diff
@@ -53,10 +53,19 @@ export function activate(context: vscode.ExtensionContext) {
     //register a link provider for this extension's "BrightScript Log" output
     vscode.languages.registerDocumentLinkProvider({ language: 'Log' }, docLinkProvider);
     //give the launch config to the link provder any time we launch the app
-    vscode.debug.onDidReceiveDebugSessionCustomEvent((e) => {
+    vscode.debug.onDidReceiveDebugSessionCustomEvent(async (e) => {
         if (e.event === 'BSLaunchStartEvent') {
             docLinkProvider.setLaunchConfig(e.body);
             logOutputManager.setLaunchConfig(e.body);
+        } else if (!e.event) {
```
georgejecook: nice touch :)

chrisdp: Thanks!

src/BrightScriptDebugSession.ts
```diff
@@ -281,15 +282,25 @@ export class BrightScriptDebugSession extends DebugSession {
             await this.rokuAdapter.activate();
 
             if (!error) {
-                console.log(`deployed to Roku@${args.host}`);
-                this.sendResponse(response);
+                if (this.rokuAdapter.connected) {
+                    // Host connection was established befor the main public process was completed
```
TwitchBronBron: `befor` should be `before`

src/BrightScriptDebugSession.ts
```diff
@@ -319,6 +330,14 @@ export class BrightScriptDebugSession extends DebugSession {
         }
     }
 
+    /**
+     * Reusable function just to keep things DRY
+     */
+    private successfulDeployPreparations(response: DebugProtocol.LaunchResponse, host: string) {
```
TwitchBronBron: I don't think the abstraction is worth the effort in this case. This function is only called in 2 places, and there are only 2 lines in this function, so it's a little more readable to just inline them and remove this function.

src/BrightScriptDebugSession.ts
```diff
@@ -810,6 +829,10 @@ export class BrightScriptDebugSession extends DebugSession {
         this.sendResponse(response);
     }
 
+    /**
+     * Creates and registers the main events for the RokuAdapter
+     * @param {string} host ip address to connect to
```
TwitchBronBron: for typescript, you don't need to add the data type `{string}` in jsdoc comments because the type is already inferred.

src/extension.ts
TwitchBronBron: What happens when `e.event` is truthy, but is not `"BSLaunchStartEvent"`?

chrisdp: In that use case it could be ether a `BSLogOutputEvent` or a `BSRendezvousEvent` once once #169  is merged.

PR #181, opened 2019-07-23
Auto name component libraries

chrisdp:


src/util.ts
```diff
@@ -76,6 +76,31 @@ export async function fileExists(filePath: string) {
     });
 }
 
+/**
+ * Reads the the manifest file and converts to a javascript object skipping empty lines and comments
+ * @param path location of the manifest file
+ */
+export async function convertManifestToObject(path: string): Promise<{ [key: string]: string } | undefined> {
```
TwitchBronBron: You should consider using the manifest parser from `brs` (see [this link](https://github.com/TwitchBronBron/brightscript-language/blob/ff3ab1060c6f3bd840f7b7fcb14eb1a20c778357/src/files/BrsFile.ts#L94)) instead of rolling your own.

chrisdp: The act of converting to use `brs` for this was simple but was blocked do to the fact that values like this: 
`build_version=00001` become `[build_version, 1]`

As a result I am expecting: 
`SampleComponentLib_1_1_00001.zip` but get `SampleComponentLib_1_1_1.zip`

TwitchBronBron: Ok, that's fine. Then can you add a test or two for `convertManifestToObject`?

src/util.ts
```diff
@@ -76,6 +76,31 @@ export async function fileExists(filePath: string) {
...
+    return new Promise(async (resolve) => {
```
TwitchBronBron: Since this is already an async function, you don't need to wrap this in a promise. Just `return undefined`, or `return manifestValues`.

PR #182, opened 2019-07-23
Tracker task injection on launch

chrisdp:
The goal of this PR is to allow users to manage their version of the tracker task across all channels. This way the tracker task code is not committed to their channel repo and so that when the tracker task and desktop RALE app is updated they only need to update it in the one location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,20 +270,20 @@ There are several string placeholders you can use when defining your deep link u
 
 ## RALE Support
 You can also have the extension automatically inject the `TrackerTack.xml` and the code snippet required to start the tracker task.
-Do do this you need a few simple things:
+To do this you need a few simple things:
 - In your VS Code user settings add the `brightscript.rokuAdvancedLayoutEditor.trackerTaskFileLocation` setting. (See [Extension Settings](#Extension-Settings) for more information)
-- Add the entry point comment `' vs_code_tracker_entry` to your code.
+- Add the entry point comment `' vscode_rale_tracker_entry` to your code.
   - This is optional as you can still include the the code to create the tracker task your self.
-  - I recommend adding it to the end of your `screen.show()` call. For example: `screen.show() ' vs_code_tracker_entry`
+  - I recommend adding it to the end of your `screen.show()` call. For example: `screen.show() ' vscode_rale_tracker_entry`
   - This can be added anywhere in the channel including source files but it must be on or after the your call to `screen.show()`
-- Set the `injectTrackerTask` value to true in your `launch.json`. For example:
+- Set the `injectRaleTrackerTask` value to true in your `launch.json`. For example:
 
 ```json
 {
     "type": "brightscript",
     "rootDir": "${workspaceFolder}/dist",
     "host": "192.168.1.2",
-    "injectTrackerTask": true
+    "injectRaleTrackerTask": true
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ A VSCode extension to support Roku's BrightScript language.
   - logpoints
   - hit count breakpoints
 - Automatic Rendezvous tracking when `logrendezvous` is enabled on the Roku. See [here](https://developer.roku.com/docs/developer-program/debugging/debugging-channels.md#scenegraph-debug-server-port-8080-commands) for information on how to enable rendezvous logging your Roku.
+- Injection of the Roku Advanced Layout Editor(RALE) task from a single user managed version
+  - This helps avoid committing the tracker to you repo and also lets you manage what version you want installed rather then other users on the project
+  - See ([Extension Settings](#Extension-Settings) and [RALE Support](#RALE-Support) for more information)
 - Publish directly to a roku device from VSCode (provided by [roku-deploy](https://github.com/TwitchBronBron/roku-deploy))
   - Also supports zip'ing and static file hosting for Component Libraries ([click here](#Component-Libraries) for more information)
 - Basic symbol navigation for document and workspace ("APPLE/Ctrl + SHIFT + O" for document, "APPLE/Ctrl + T" for workspace)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -264,6 +285,7 @@ This extension contributes the following settings:
 * `brightscript.output.hyperlinkFormat`: specifies the display format for log output `pkg` link
 * `brightscript.deviceDiscovery.showInfoMessages`: If set to true, an info toast will be shown when a Roku device has been found on the network.
 * `brightscript.deviceDiscovery.enabled`: If set to true, the extension will automatically watch and scan the network for online Roku devices. This can be pared with the `${promptForHost}` option in the launch config to display a list of online Rokus, removing the need to constantly change the host IP in your config files.
+* `brightscript.rokuAdvancedLayoutEditor.trackerTaskFileLocation`: This is an absolute path to the TrackerTask.xml file to be injected into your Roku channel during a debug session. (i.e. `/Users/user/roku/TrackerTask/TrackerTask.xml`)
```
TwitchBronBron: Is there a reason this needs to be forced to absolute path? Why can't we support workspace-relative paths?

package.json
```diff
@@ -548,6 +553,11 @@
                     "default": true,
                     "description": "If set to true, the extension will automatically watch and scan the network for online Roku devices. This can be pared with the ${promptForHost} option in the launch config to display a list of online Rokus, removing the need to constantly change the host IP in your config files.",
                     "scope": "resource"
+                },
+                "brightscript.rokuAdvancedLayoutEditor.trackerTaskFileLocation": {
```
TwitchBronBron: I understand wanting this to be a user setting, and that's fine to support it. However, we may want to consider supporting this as a launch config setting as well. The launch.json setting would override this value if found.

src/BrightScriptDebugSession.ts
```diff
@@ -948,6 +959,59 @@ export class BrightScriptDebugSession extends DebugSession {
         await Promise.all(promises);
     }
 
+    /**
+     * Will search the project files for the comment "\' vs_code_tracker_entry" and replace it with the code needed to start the TrackerTask.
+     * @param stagingPath
+     */
+    public async injectTrackerTaskCode(stagingPath: string) {
```
TwitchBronBron: Can you add a unit test to validate this works?

package.json
chrisdp: The reason this is a user setting is because this is workspace independent. I'd be happy to hope on a short call to talk over this feature.

README.md
chrisdp: The reason this is a absolute path is because this is workspace independent. I'd be happy to hope on a short call to talk over this feature.

src/BrightScriptDebugSession.ts
chrisdp: Tests added.

README.md
```diff
@@ -247,6 +250,24 @@ There are several string placeholders you can use when defining your deep link u
 
  - `${promptForDeepLinkUrl}` - if the entire `deepLinkUrl` is set to this, then at debug launch time, an input box will appear asking you to input the full deep link url.
 
+## RALE Support
+You can also have the extension automatically inject the `TrackerTack.xml` and the code snippet required to start the tracker task.
+Do do this you need a few simple things:
```
TwitchBronBron: Spelling. Says `Do do`. Should say `To do`.

package.json
```diff
@@ -308,6 +308,11 @@
                                     "normal"
                                 ]
                             },
+                            "injectTrackerTask": {
```
TwitchBronBron: I think we should include RALE somewhere in this name. something like `injectRaleTrackerTask`? Thoughts?

src/BrightScriptDebugSession.ts
```diff
@@ -989,6 +1000,59 @@ export class BrightScriptDebugSession extends DebugSession {
...
+        // Search for the tracker task entry injection point
+        let trackerEntryTerm = `('\\s*vs_code_tracker_entry[^\\S\\r\\n]*)`;
```
TwitchBronBron: Let's make `vs_code` one word `vscode`. 

We should probably use the world `rale` somewhere in this key? `'vscode_rale_tracker_entry`. Thoughts?

TwitchBronBron:
@chrisdp whatever key we finally land on for the comment (`vscode_rale_tracker_entry` for example), we should add it to the language grammar so it appears a slightly different color in the editor. That is in the `syntaxes/brightscript.tmLanguage.json` file.

src/BrightScriptDebugSession.spec.ts
```diff
@@ -415,6 +416,116 @@ describe('Debugger', () => {
...
+    });
+
+    describe('injectRaleTrackerTaskCode', () => {
+        let key: string;
+        let trackerTaskCode: string;
+        let folder;
+
+        beforeEach(() => {
+            key = 'vscode_rale_tracker_entry';
+            trackerTaskCode = `if true = CreateObject("roAppInfo").IsDev() then m.vs_code_tracker_task = createObject("roSGNode", "TrackerTask") ' Roku Advanced Layout Editor Support`;
```
TwitchBronBron: Your test has the old `vs_code_tracker_task` text, but it's not failing the test. Is that a bug in the test?

src/BrightScriptDebugSession.ts
```diff
@@ -1101,6 +1112,59 @@ export class BrightScriptDebugSession extends DebugSession {
...
+
+        let keys = Object.keys(results);
+        if (keys.length === 0) {
+            // Do not throw an error as we don't want to prevent the user from launching the channel
+            // just because they don't have a local version of the TrackerTask.
+            this.sendDebugLogLine('WARNING: Unable to find an entry point for Tracker Task.');
+            this.sendDebugLogLine('Please make sure that you have the following comment in your BrightScript project: "\' vscode_rale_tracker_entry"');
+        } else {
+            // This code will start the tracker task in the project
+            let trackerTaskSupportCode = `if true = CreateObject("roAppInfo").IsDev() then m.vs_code_tracker_task = createObject("roSGNode", "TrackerTask") ' Roku Advanced Layout Editor Support`;
```
TwitchBronBron: This is using the old `vs_code_tracker_task` but the test is not failing.

src/BrightScriptDebugSession.spec.ts
chrisdp: No, what you'er looking at is the code that is going to be injected into the channel in the place of the comment. It's creating a `m.vs_code_tracker_task` variable to store the tracker in. I can change this variable to `m.vscode_rale_tracker_task` if you would like.

src/BrightScriptDebugSession.ts
chrisdp: No, what you'er looking at is the code that is going to be injected into the channel in the place of the comment. It's creating a `m.vs_code_tracker_task` variable to store the tracker in. I can change this variable to `m.vscode_rale_tracker_task` if you would like.

src/BrightScriptDebugSession.spec.ts
TwitchBronBron: ah. ok, that makes more sense. Yeah, let's standardize it to `vscode` anywhere that `vs_code` is used.

chrisdp: Done.

PR #183, opened 2019-07-24
Bs_const support in launch configs

chrisdp:
Allows users to change existing bs_consts in the channel manifest to change channel behaviour or features with the launch configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ example config:
     "type": "brightscript",
     "rootDir": "${workspaceFolder}/dist",
     "host": "192.168.1.2",
-    "bs_const": {
+    "bsConst": {
         "debug": true,
         "logging": false
     }
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ A VSCode extension to support Roku's BrightScript language.
   - LogLevel (example `^\[(info|warn|debug\]`)
   - Include (example `NameOfSomeInterestingComponent`)
   - Exclude (example `NameOfSomeNoisyComponent`)
+- Variable `bs_const` values using the `launch.json` (see the [BS_Const](#BS_Const) section for more information)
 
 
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -164,6 +165,23 @@ If you have a custom build process that pulls in files from multiple source dire
 }
 ```
 
+## BS_Const
+
+If you use `bs_const` in your project manifest you can define septate launch configs in your `launch.json` allowing for easy changing without modifying the manifest your self. This helps prevent accidentally committing a change to the `bs_consts` in your project. You can not define a constant that is not also in your manifest. See the [Manifest constant](https://developer.roku.com/en-ca/docs/references/brightscript/language/conditional-compilation.md#manifest-constant) documentation for more info on there format.
```
TwitchBronBron: spelling mistake. `septate` should be `separate`.

TwitchBronBron: Also, `your self` should be `yourself`.

TwitchBronBron: Also, `more info on there format`, `there` should be `their`

src/BrightScriptDebugSession.ts
```diff
@@ -319,6 +326,56 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                    } else {
+                        results.push(key);
+                    }
+
+                    return results;
+                }, []);
+
+                // check for consts that where not in the manifest
+                if (missingConsts.length > 0) {
+                    throw new Error(`The following bs_const keys were not defined in the channels manifest:\n\n${missingConsts.join(',\n')}`);
```
TwitchBronBron: Spelling: `channels` should be possessive  `channel's`.

src/BrightScriptDebugSession.ts
```diff
@@ -319,6 +326,56 @@ export class BrightScriptDebugSession extends DebugSession {
...
+                } else {
+                    // update the manifest and write to the staging folder
+                    fileContents = fileContents.replace(bsConstLine, newLine);
+                    await fsExtra.writeFile(manifestPath, fileContents);
+                }
+            } else {
+                throw new Error('bs_const was defined in the launch.json but not in the channels manifest');
```
TwitchBronBron: Spelling: channels should be possessive channel's.

TwitchBronBron:
Well implemented! Can you please write a few unit tests to verify that the `updateManifestBsConsts` function works properly? I would modify the `updateManifestBsConsts` function so that it accepts the consts and the CONTENTS of the file. So that way, the unit test is much easier to test because you can pass manifest contents in to the function to test it. I added a few comments at the relevant locations.

README.md
TwitchBronBron: It looks like these are still unresolved?

PR #184, opened 2019-07-26
takes sourceDirs into account when resolving the method name, for the FilenameAndFunction log setting

georgejecook:
note - this will only work for 1 source dir we should iterate over more but; 1. I don't have time right now and 2; not sure about performance ramifications of that. I'll raise an issue to capture this though.


TwitchBronBron:
@georgejecook There's a bunch of `brighterscript` changes in this PR. Those need to be taken out before I can approve this.

georgejecook:
ha! yes. I thought I'd deleted this pr - sorry for wasting your time @TwitchBronBron I've not yet got a reasonable solution to this problem.

PR #212, opened 2019-10-17
Adding a setting to optionally remove interface text after completion item is selected

Romans-I-XVI:
I like the setup of needing to explicitly call interfaces to trigger completion items but I don't usually want all that explicit interface calling in my final code. I was manually going back and removing it so I decided to add an option in the extension to do this automatically.


TwitchBronBron:
Hey, while I understand the motivation for this, I don't think it's something we want to include in the extension. This is a pretty rare case, and is something that's fairly simple to implement as a one-off node script/vscode task for the few people who want to use it.  

Thanks though.

PR #213, opened 2019-10-21
Fixed an issue as of Roku 9.2 where the paths to component library fi…

chrisdp:
…les could not be resolved.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,5 +12,10 @@
     "editor.insertSpaces": true,
     "editor.detectIndentation": false,
     "editor.tabSize": 4,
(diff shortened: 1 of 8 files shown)


.vscode/settings.json
```diff
@@ -12,5 +12,10 @@
     "editor.insertSpaces": true,
     "editor.detectIndentation": false,
     "editor.tabSize": 4,
-    "typescript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false
+    "typescript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false,
+    "workbench.colorCustomizations": {
```
TwitchBronBron: @chrisdp  I forgot to mention this earlier, so I approved it so I could start the release. But this shouldn't have been committed.

PR #215, opened 2019-12-18
Adds some logic to the configuration provider to default config value…

chrisdp:
…s with missing env values to the extension default values.


TwitchBronBron:
@chrisdp can you take a look at the failing tests and get them to pass? https://travis-ci.org/rokucommunity/vscode-brightscript-language/builds/626797825?utm_source=github_status&utm_medium=notification

PR #217, opened 2020-01-21
Optimize component library file renaming

triwav:
Optimize component library file renaming by only running the regex once on every brs file appending the comp lib postfix


TwitchBronBron:
Looks good. Could you also put together a pull request for the master branch with the same fix? I'm sure the non-v2.beta users would also enjoy this.

PR #218, opened 2020-02-12
Speed up RALE insertion

triwav:
Speed up RALE insertion by using replace in file instead and doing non regex based lookup


TwitchBronBron:
@triwav, the CI builds have a failing test in all 3 of the target operating systems. Can you please fix the test?

PR #224, opened 2020-04-06
Update README.md

timalacey:
`brsconfig.json` has no effect, `bsconfig.json` does


TwitchBronBron:
@timalacey I thought for sure that I had changed these in the v2 branch before we merged into master, but I'm either crazy or they got messed up during the merge. In any case, these needed corrected. Thanks a bunch!

PR #241, opened 2020-05-18
feat(DeclarationProvider): Adds temporary support for brighterscript symbol detection and navigation

georgejecook:
I've been using this code for several months now; it's just as shoe-string as the rest of the impl; but it does get method feedback and navigation support for common bs use cases, like namespaces, extends, classes, and constructors.

### DIFF
--- a/src/DeclarationProvider.ts
+++ b/src/DeclarationProvider.ts
@@ -73,7 +73,7 @@ export class DeclarationProvider implements Disposable {
     constructor() {
         const subscriptions: Disposable[] = [];
 
-        const watcher = vscode.workspace.createFileSystemWatcher('**/*.brs');
+        const watcher = vscode.workspace.createFileSystemWatcher('**/*.{brs,bs}');
         watcher.onDidCreate(this.onDidChangeFile, this);
         watcher.onDidChange(this.onDidChangeFile, this);
         watcher.onDidDelete(this.onDidDeleteFile, this);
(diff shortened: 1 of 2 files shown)


src/DeclarationProvider.ts
```diff
@@ -190,14 +194,41 @@ export class DeclarationProvider implements Disposable {
...
+        let classSymbol: BrightScriptDeclaration | null;
 
         for (const [line, text] of iterlines(input)) {
             // console.log("" + line + ": " + text);
             funcEndLine = line;
             funcEndChar = text.length;
 
             //FUNCTION START
-            let match = /^\s*(?:function|sub)\s+(.*[^\(])\s*\((.*)\)/i.exec(text);
+            let match = /^\s*(?:override)*\s*(?:public|private)*\s*(?:function|sub)\s+(.*[^\(])\s*\((.*)\)/i.exec(text);
```
TwitchBronBron, line 231: @georgejecook 

The syntax should be (pseudocode):
optional AccessModifier (`public`|`protected`|`private`)
optional `override`
`sub`|`function`
FunctionName

Examples:
```
public override sub GetName()
protected override sub GetName()
private override sub GetName()
public sub GetName()
protected sub GetName()
private sub GetName()
sub GetName()
```

src/DeclarationProvider.ts
```diff
@@ -231,11 +263,11 @@ export class DeclarationProvider implements Disposable {
                 continue;
             }
 
-            // //VAR
-            match = /^\s*(?:m\.)([a-zA-Z_0-9]*)/i.exec(text);
+            // //FIELD
+            match = /^(?!.*\()(?: |\t)*(public|private)(?: |\t)*([a-z|\.|_]*).*((?: |\t)*=(?: |\t)*.*)*$/i.exec(text);
```
TwitchBronBron, line 267: Missed `protected` here.

georgejecook: oh. I didn't realize fields could be protected. thanks

georgejecook: addressed

georgejecook:
@TwitchBronBron sorry - had to force push this- I accidentally pushed a settings file up..

TwitchBronBron:
> @TwitchBronBron sorry - had to force push this- I accidentally pushed a settings file up..

lol you can always just delete it in a subsequent commit. If you add then commit, then delete and commit, it's as if the file never existed.

TwitchBronBron:
Looks good, thanks for making the changes.

PR #248, opened 2020-05-22
Adds basic snippets for brs/bs and xml files

markwpearce:
Adds some basic snippets for Roku development.

For example in .bs/.brs files:

1. Typing `sub<tab>` generates:
```brs
sub subName()

end sub
```
with "subName" editable

2. Typing `for-each<tab>`generates:
```brs
for each item in collection

end for
```
with "item" and "collection" editable

For example in .xml files:

1. Typing `<component<tab>` generates
 ```xml
<component name="ComponentName">
  <script type="text/brightscript" uri="ComponentName.brs" />

</component
```

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -584,6 +584,18 @@
             {
                 "language": "brightscript",
                 "path": "./snippets.json"
+            },
```
markwpearce, line 587: Note:
I wanted to have different snippets for brightscript and brighterscript, but I think because they are actually defined to use the same grammar, the snippets show up in both places. To be more clear, I just set them all to brightscript, although  kept the brighterscript specific ones in their own file

snippets/brightscript.json
```diff
@@ -0,0 +1,71 @@
...
+            "\t$0",
+            "end for"
+        ],
+        "description": "For Each item in a collection"
+    },
+
+    "Function": {
+        "prefix": "function",
+        "body": [
+            "function ${1:functionName}()",
```
triwav, line 23: Should we have a placeholder for params?

snippets/brighterscript.json
```diff
@@ -0,0 +1,44 @@
+{
+    "Class": {
+        "prefix": "class",
+        "body": [
+            "class ${1:ClassName}",
+            "\tsub new()",
```
chrisdp, line 6: Humm is there a way to provide this matching the user/workspace setting for indentation? Or with VS Code already handle this when injecting the snippet?

snippets/brightscript.json
markwpearce: Yes, we should...

chrisdp:
We should see if we can at least separate the BS only snippets from the BRS once if we can. @TwitchBronBron ideas for this?

snippets/brighterscript.json
markwpearce: `\t` in the snippet body matches whatever your workspace indentation settings are.

snippets/brightscript.json
markwpearce: I updated the PR with params placeholder in functions/subs

georgejecook:
what's the status of this? waiting on separating out snippets for bs/brs?

markwpearce:
@georgejecook I don't think there's anything else that I can do. It's a bigger change to modify the grammars so they are perceived differently by vs code

package.json
TwitchBronBron: That's fine for now. We can always split them out later. Thanks for keeping them in separate files.

PR #256, opened 2020-06-12
Enables replacement of all ${env:VARIABLE} replacements

jfrux:
Now handles all env variable replacements for launch.json configs.

Even if you use them in your `bsconfig.json` file they will replace - which is a nice side effect for some of the projects I'm part of right now actually.

I left the existing piece that was setting default values based on config keys that you put in the env file but I'm not sure how applicable that is so you can let me know if that needs to just get scrapped now.

Let me know if anything sounds like it might break something else but it appears to be working as intended as of right now.

### DIFF
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -295,22 +295,6 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
             }
 
             config = JSON.parse(configString);
-
-            // apply any default values to env placeholders
-            for (let key in config) {
-                let configValue = config[key];
-                let match: RegExpMatchArray;
-                //replace all environment variable placeholders with their values
-                while (match = regexp.exec(configValue)) {
-                    let environmentVariableName = match[1];
-                    let environmentVariableValue = envConfig[environmentVariableName];
-                    if (!environmentVariableValue) {
-                        configValue = this.configDefaults[key];
-                        console.log(`The configuration value for ${key} was not found in the env file under the name ${environmentVariableName}. Defaulting the value to: ${configValue}`);
-                    }
-                }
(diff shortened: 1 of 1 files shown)


TwitchBronBron:
Looks great. Concerning the "existing piece that was setting default values based on config keys", it seems like that does the same thing that your code does, so if you're pretty confident yours will catch all of the env substitutions, then we can scrap that piece. So yeah, **go ahead and remove that piece and then I'll merge the PR**.

Also an unrelated side-note: if you put env settings in your bsconfig, you then can't run the `bsc` tool to build the code from the command prompt using that bsconfig.json. Not a big deal if you only use bsconfig.json for the extension, but just wanted to mention it just in case.

jfrux:
> Looks great. Concerning the "existing piece that was setting default values based on config keys", it seems like that does the same thing that your code does, so if you're pretty confident yours will catch all of the env substitutions, then we can scrap that piece. So yeah, **go ahead and remove that piece and then I'll merge the PR**.
> 
> Also an unrelated side-note: if you put env settings in your bsconfig, you then can't run the `bsc` tool to build the code from the command prompt using that bsconfig.json. Not a big deal if you only use bsconfig.json for the extension, but just wanted to mention it just in case.

It's not that I think they do the same thing per say... just not sure it does anything at all...
Are there "default" env variables you would put in place that woul dneed to be replaced even if the `.env` file doesn't contain them? Possibly `ROKU_USER` or something?

jfrux:
@TwitchBronBron Yeah, we need to have the defaults for your tests to pass.

jfrux:
So I ended up leaving it in... feel free to merge whenever you're ready.

jfrux:
@TwitchBronBron Any last concerns to get this merged?

PR #262, opened 2020-07-09
Adds support for logouput format options for link:// style links

georgejecook:
Wouldn't mind a preliminary look at this; if we really need, I can add some additional unit tests; but I know it's not borked, coz I'm using it heavily

### DIFF
--- a/src/LogDocumentLinkProvider.ts
+++ b/src/LogDocumentLinkProvider.ts
@@ -81,29 +81,27 @@ export class LogDocumentLinkProvider implements vscode.DocumentLinkProvider {
         return this.fileMaps[pkgPath];
     }
 
-    public addCustomLink(customLink: CustomDocumentLink, isFilePath: boolean = false) {
-        if (isFilePath) {
-            let range = new Range(new Position(customLink.outputLine, customLink.startChar), new Position(customLink.outputLine, customLink.startChar + customLink.length));
-            let uri = vscode.Uri.file(customLink.pkgPath);
+    public addCustomFileLink(customLink: CustomDocumentLink) {
+        let range = new Range(new Position(customLink.outputLine, customLink.startChar), new Position(customLink.outputLine, customLink.startChar + customLink.length));
+        let uri = vscode.Uri.file(customLink.pkgPath);
+        if (customLink.lineNumber) {
+            uri = uri.with({ fragment: customLink.lineNumber.toString().trim() });
+        }
+
+        this.customLinks.push(new DocumentLink(range, uri));
+    }
+
+    public addCustomPkgLink(customLink: CustomDocumentLink) {
+        let fileMap = this.getFileMap(customLink.pkgPath);
+        if (fileMap) {
+            let uri = vscode.Uri.file(fileMap.src);
             if (customLink.lineNumber) {
                 uri = uri.with({ fragment: customLink.lineNumber.toString().trim() });
             }
-
+            let range = new Range(new Position(customLink.outputLine, customLink.startChar), new Position(customLink.outputLine, customLink.startChar + customLink.length));
             this.customLinks.push(new DocumentLink(range, uri));
-
         } else {
-
-            let fileMap = this.getFileMap(customLink.pkgPath);
-            if (fileMap) {
-                let uri = vscode.Uri.file(fileMap.src);
-                if (customLink.lineNumber) {
-                    uri = uri.with({ fragment: customLink.lineNumber.toString().trim() });
-                }
-                let range = new Range(new Position(customLink.outputLine, customLink.startChar), new Position(customLink.outputLine, customLink.startChar + customLink.length));
-                this.customLinks.push(new DocumentLink(range, uri));
-            } else {
-                console.log('could not find matching file for link with path ' + customLink.pkgPath);
-            }
+            console.log('could not find matching file for link with path ' + customLink.pkgPath);
         }
     }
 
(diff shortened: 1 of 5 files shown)


TwitchBronBron:
Looks good to me. Just add a few unit tests to validate the new supported path types, and .bs files, and then I'll be ready to merge it.

georgejecook:
note; I've added some more tests. However - I want to test this on a non wps setup; as my line numbers are not linking to the right place, and I can't see why - think it's my wps setup. I'll try tomorrow am on my linux machine and report back if this is ready for review.

georgejecook:
@TwitchBronBron I've updated this pr, added more tests, and made another minor improvement; been using this branch all week.

src/LogDocumentLinkProvider.ts
```diff
@@ -78,19 +81,29 @@ export class LogDocumentLinkProvider implements vscode.DocumentLinkProvider {
         return this.fileMaps[pkgPath];
     }
 
-    public addCustomLink(customLink: CustomDocumentLink) {
-        let match: RegExpExecArray;
-
-        let fileMap = this.getFileMap(customLink.pkgPath);
-        if (fileMap) {
-            let uri = vscode.Uri.file(fileMap.src);
+    public addCustomLink(customLink: CustomDocumentLink, isFilePath: boolean = false) {
```
TwitchBronBron, line 84: This function doesn't share any code whatsoever between isFilePath: true | false, ...so how about we break these into two functions:
 - `addCustomPkgLink`
 - `addCustomFileLink`

src/LogDocumentLinkProvider.ts
```diff
@@ -99,13 +112,10 @@ export class LogDocumentLinkProvider implements vscode.DocumentLinkProvider {
...
-        //remove any preceding file scheme
-        pkgPath = util.removeFileScheme(pkgPath);
-
-        //use debugRootDir if provided, or rootDir if not provided.
-        let rootDir = this.launchConfig.debugRootDir ? this.launchConfig.debugRootDir : this.launchConfig.rootDir;
-
-        let clientPath = path.normalize(path.join(rootDir, pkgPath));
-        return clientPath;
+        let mappedPath = this.getFileMap(pkgPath);
+        if (!mappedPath) {
```
TwitchBronBron, line 116: Can you add a few comments to the statements in this function? It's not clear which situations would cause a `mappedPath` to be undefined, and it's also not clear why we need to try looking for the `getAlternateBrsFileName` version of the `pkgPath` either.

src/LogOutputManager.ts
```diff
@@ -39,7 +39,7 @@ export class LogOutputManager {
         this.includeRegex = null;
         this.logLevelRegex = null;
         this.excludeRegex = null;
-        this.pkgRegex = /(\w+:\/.*\.(?:brs|xml))[ \t]*(?:\((\d+)(?:\:(\d+))?\))?/;
+        this.pkgRegex = /(?:\s*at line (\d*) of file )*(?:(pkg:\/|file:\/\/)(.*\.(bs|brs|xml)))[ \t]*(?:(?:(?:\()(\d+)(?:\:(\d+))?\)?)|(?:\:(\d+)?))*/;
```
TwitchBronBron, line 42: Can you add a comment to this regex explaining the various formats that are supported? It's not clear why `at line (\d*) of file)` is included here. Perhaps a block comment with examples of what log formats will be captured by this regex.

src/LogOutputManager.ts
```diff
@@ -258,13 +258,18 @@ export class LogOutputManager {
...
             let match = this.pkgRegex.exec(logLine.text);
             if (match) {
-                const pkgPath = match[1];
-                const lineNumber = Number(match[2]);
-                const filename = this.getFilename(pkgPath);
-                const extension = pkgPath.substring(pkgPath.length - 4);
-                let customText = this.getCustomLogText(pkgPath, filename, extension, Number(lineNumber), logLineNumber);
-                const customLink = new CustomDocumentLink(logLineNumber, match.index, customText.length, pkgPath, lineNumber, filename);
-                this.docLinkProvider.addCustomLink(customLink);
+                const isFilePath = match[2] === 'file://';
```
TwitchBronBron, line 261: the file protocol is more complicated than this. Sometimes it can be `file:/`, `file://`, `file:///`, `file:////`. (See [how many slashes on wikipedia](https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?).  We should be using a tool to handle this. [file-url](https://github.com/sindresorhus/file-url) handles parsing things into file urls, there's also the vscode Uri class that takes a uri and provides the various parts (like scheme, path ,etc...).  

at the very least, perhaps consider a regular expression match on `match[2]` instead. 
```
match[2].match(
    /file:\/{1,4}/
)
```

georgejecook: this is a moot point. note the origina regex; it only catches 2 slashes, which are matched here the rest is the path.
that will always be valid, even on windows, as per the wiki link you shared states: "Note that the drive letter followed by a colon and slash is part of the acceptable file URI."

so we don't need to do anything here; did you experience an issue with this?

georgejecook:
thanks for the feedback, @TwitchBronBron ; made the changes you asked, save for file regex, which I don't believe needs any action. Try it and let me know if I'm wrong  - I'll change it if I have to :)

src/LogOutputManager.ts
TwitchBronBron: @georgejecook btw, the comments you added are perfect. Thanks!

PR #269, opened 2020-07-28
bsfmt.json

### LINKED ISSUE #270: Support bsfmt.json files natively within the extension
The extension should support loading `bsfmt.json` formatter options from brighterscript-formatter.

TwitchBronBron:
Adds support for loading bsfmt.json formatter options. If `bsfmt.json` exists, all formatter-related user/workspace settings will be ignored. 

Fixes #270.

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -32,10 +32,8 @@ export class Formatter implements DocumentRangeFormattingEditProvider {
             let text = document.getText();
             let formatter = new BrighterScriptFormatter();
             let formattedText = formatter.format(text, <FormattingOptions>{
-                //override with user settings
-                ...userSettingsOptions,
-                //override user settings with bsfmt.json options
-                ...(bsfmtOptions ?? {}),
+                //if we found bsfmt.json options, use ONLY those. Otherwise, use any options found from user/workspace settings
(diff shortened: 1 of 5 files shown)


TwitchBronBron:
@elsassph could you take a look at this to verify it suits your needs?

src/formatter.ts
```diff
@@ -15,13 +16,24 @@ import * as vscode from 'vscode';
 export class Formatter implements DocumentRangeFormattingEditProvider {
 
     public provideDocumentRangeFormattingEdits(document: TextDocument, range: Range, options: vscode.FormattingOptions): ProviderResult<TextEdit[]> {
-        let config = workspace.getConfiguration('brightscript.format');
+
+        //TODO is there anything we can to do to better detect when the same file is used in multiple workspaces?
+        //vscode seems to pick the lowest workspace (or perhaps the last workspace?)
+        const workspaceFolder = workspace.getWorkspaceFolder(document.uri);
+
+        let bsfmtOptions = new Runner().getBsfmtOptions({
```
elsassph, line 24: Does that actually return `undefined/null` if no config is present?

elsassph: Never mind, I checked and that's the case.

PR #279, opened 2020-10-14
Use language server for features when available

triwav:


src/LanguageServerManager.ts
```diff
@@ -16,7 +16,7 @@ import { CustomCommands } from 'brighterscript';
 import { CodeWithSourceMap } from 'source-map';
 import { Deferred } from 'brighterscript';
 
-export class LanaguageServerManager {
```
TwitchBronBron, line 19: Nice catch!

src/extension.ts
```diff
@@ -107,14 +107,16 @@ export class Extension {
...
 
-        // experimental placeholder
-        // context.subscriptions.push(vscode.languages.registerCompletionItemProvider(selector, new BrightScriptCompletionItemProvider(), '.'));
-        context.subscriptions.push(vscode.languages.registerDefinitionProvider(selector, definitionProvider));
-        context.subscriptions.push(vscode.languages.registerDocumentSymbolProvider(selector, new BrightScriptDocumentSymbolProvider(declarationProvider)));
-        context.subscriptions.push(vscode.languages.registerWorkspaceSymbolProvider(new BrightScriptWorkspaceSymbolProvider(declarationProvider, symbolInformationRepository)));
-        context.subscriptions.push(declarationProvider);
-        vscode.languages.registerReferenceProvider(selector, new BrightScriptReferenceProvider());
-        vscode.languages.registerSignatureHelpProvider(selector, new BrightScriptSignatureHelpProvider(definitionRepo), '(', ',');
+        if (!languageServerManager.isLanguageServerEnabledInSettings) {
```
TwitchBronBron, line 110: Users can toggle the language server on and off by switching the user setting, and it works immediately, without needing to restart vscode. By doing these subscriptions in `extension.activate()`, it means anytime the user changes that setting, they will need to restart vscode. 

We should move these registrations into the `LanguageServerManager`, and activate/deactivate these every time the `enableLanguageServer` option is changed, on the fly, without needing a vscode restart. 

I believe you can get the subscriptions from the `registerXYZ` function calls and keep them around for when you want to unsubscribe.

TwitchBronBron (approved):
Looks great!

PR #302, opened 2021-04-14
Added new showOutputPanelOnStartup setting

chrisdp:


package.json
```diff
@@ -1387,6 +1387,12 @@
                     "default": false,
                     "scope": "resource"
                 },
+                "brightscript.showOutputPanelOnStartup": {
```
TwitchBronBron, line 1390: I think it might be better for us to deprecate `focusOutputPanelOnStartup`, and instead add a `outputPanelStartupBehavior` option with values of `"show" | "focus" "nothing"`.

Also, this should be supported in the launch configuration as well, but you don't have to do that work here...we could add support for that later.

PR #316, opened 2021-06-24
Syntax highlighting for embedded brs in scenegraph

TwitchBronBron:
This adds syntax highlighting support for embedded brightscript/brighterscript code in CDATA tags within scenegraph xml files.

Before:
image

After:
image

**NOTE:** this might prevent xml formatters from working properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
         "clean": "rimraf dist",
         "install-local": "node scripts/install-local.js",
         "install-pull-local": "node scripts/install-local.js --pull",
-        "uninstall-local": "node scripts/uninstall-local.js"
+        "uninstall-local": "node scripts/uninstall-local.js",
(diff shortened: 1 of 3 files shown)


elsassph:
I concur with: do not break XML coding experience.

chrisdp:
Looping back to this, what was the reason this could break other tools? Did this effectively change the file type in the editor?

TwitchBronBron:
> Looping back to this, what was the reason this could break other tools? Did this effectively change the file type in the editor?

The problem is that we need to define our own "language" type for SceneGraph files. I believe file formatters are done based on the langauge type, not the file extension. So if you have an XML formatter, this may not work anymore.

I haven't tested it in a while though, perhaps that was wrong. 

If it does break formatting, then we will need to add our own xml formatting logic into the project before this can land.

PR #317, opened 2021-07-02
Initial RDB Implementation

triwav:
This is the initial RDB Implementation. Node tree view is still in the works and will likely be the next addition.


.vscode/settings.json
```diff
@@ -1,5 +1,8 @@
 {
-    "files.trimTrailingWhitespace": false,
+    "files.trimTrailingWhitespace": true,
+	"[markdown]": {
+		"files.trimTrailingWhitespace": false
```
TwitchBronBron, line 4: Why disable this for markdown files?

package.json
```diff
@@ -1,1699 +1,1699 @@
...
+        "validate": "svelte-check",
+        "svelte-build": "rollup -c",
+        "svelte-watch": "rollup -c -w"
+    },
+    "dependencies": {
+        "array-sort": "^1.0.0",
+        "backoff": "^2.5.0",
+        "brighterscript": "^0.37.1",
+        "brighterscript-formatter": "^1.6.0",
+        "chokidar": "^3.5.1",
```
TwitchBronBron, line 54: Should this be a devDependency? Was this to help with reloading during svelte development?

.vscode/settings.json
triwav: Markdown files use two spaces at the end of the line to represent a new line. Without this, vscode will remove those spaces.

package.json
triwav: It's only designed to be used during development to allow hot reloading of webviews built with Svelte. I did it as devDependency to make the version distributed a little smaller. Can switch it a dependency if you think we should. Was double checking we were still using it as I know we'd talked about using vscode.workspace.createFileSystemWatcher but remembered that we'd tried that but it wouldn't allow since it wasn't actually part of the workspace hence the use of chokidar.

.vscode/settings.json
TwitchBronBron: Ah, good point.

package.json
TwitchBronBron: From your comment, it feels like you believe it's already a devDependency, Based on this package.json, `chokidar` is listed as a _dependency_, and I was asking if it should be moved to a _devDependency_.

src/util.ts
```diff
@@ -167,6 +167,10 @@ class Util {
         }
         this.debounceByKey[key]();
     }
+
+    public isExtensionHostRunning() {
```
TwitchBronBron, line 171: This will be nice to have for future things as well!

tsconfig.json
```diff
@@ -8,10 +8,12 @@
         "noImplicitThis": true,
         "inlineSourceMap": false,
         "sourceMap": true,
+        "rootDir": "./src",
```
TwitchBronBron, line 11: What was the problem you are solving by adding this and the below `"include": ["src/**/*"]`.

ui/rdb/components/JSONTreeView/JSONArrow.svelte
```diff
@@ -0,0 +1,29 @@
+<script>
+    export let expanded;
+</script>
+<style>
+    .container {
+        display: inline-block;
+        cursor: pointer;
+        transform: translate(calc(0px - var(--li-identation)), 20%);
```
TwitchBronBron, line 8: Nice use of css variables!

ui/rdb/types.d.ts
```diff
@@ -0,0 +1,4 @@
+declare var acquireVsCodeApi: Function;
+declare var vscode;
```
TwitchBronBron, line 2: Is there a typescript type we could import for some of these? I thought vscode included typescript definitions? If so, we could do `import type * as vscode from 'vscode'` to make this a little safer.

ui/rdb/types.d.ts
```diff
@@ -0,0 +1,4 @@
...
+declare var requestArgsSchema;
+declare var odcCommands;
```
TwitchBronBron, line 4: Can you make all of these variables `const` instead? Also, please add types for these.

ui/rdb/views/CommandsView.svelte
```diff
@@ -0,0 +1,121 @@
+<script lang="ts">
+    window.vscode = acquireVsCodeApi();
+    import {odc} from "../ExtensionIntermediary";
+    import { commandsView } from "./CommandsView";
+
+    const storage = window.localStorage;
```
TwitchBronBron, line 6: Is there any type of cleanup we need to do with localStorage on a fresh launch of the app so we don't end up in an inconsistent/invalid state?

ui/rdb/views/CommandsView.ts
```diff
@@ -0,0 +1,45 @@
...
+                ...rawArg,
+                id: key
+            });
+        }
+        return args;
+    }
+
+    public processArgToSendToExtension(argType: string, argValue: string) {
+        if (argType == 'boolean') {
+            return argValue === 'true'
```
TwitchBronBron, line 30: What about `'false'`? Shouldn't you parse the argValue to see which it should be in this case?

ui/rdb/views/CommandsView.ts
```diff
@@ -0,0 +1,45 @@
...
+        } else if (argType == 'array' || argType == 'object') {
+            return JSON.parse(argValue);
+        } else if (argType == 'number') {
+            return Number(argValue);
```
TwitchBronBron, line 34: We probably don't the the object/boxed version of a number here. Perhaps consider just using `parseFloat` instead?

https://stackoverflow.com/questions/11988547/what-is-the-difference-between-number-and-parsefloat

tsconfig.json
triwav: If you don't limit it to the src folder it'll pull in the ts files from the ui folder and cause issues compiling. One possible alternative would be to do: 
```
"exclude": ["ui/**/*"]
```
@prajwalkshetty might know of some reason that that wouldn't work as he did most of this base setup.

ui/rdb/components/JSONTreeView/JSONArrow.svelte
triwav: Can thank @prajwalkshetty for that :)

ui/rdb/views/CommandsView.svelte
triwav: We're mostly just storing what they put in to simplify the case where they want to do the same commands over and over across launches so currently we wouldn't want to remove or cleanup any of that since it's there for that reason

ui/rdb/views/CommandsView.ts
triwav: 'false' does not equal 'true' so would convert to `false` as would anything else that does not equal 'true'. Does that make sense?

ui/rdb/views/CommandsView.ts
triwav: In that link you provided it mentions 
```
When Number is called as a function rather than as a constructor, it performs a type conversion...

Returns a Number value (not a Number object) computed by ToNumber(value) if value was supplied, else returns +0.
```
If I did `return new Number(argValue);` that would return a Number object else it returns a `number` type whether I do it the current way or `parseFloat`. Happy to change it as I realize all numbers in Javascript are floats under the hood. The Number() syntax just seemed to match better with the argType.

ui/rdb/types.d.ts
triwav: Unfortunately there isn't: https://stackoverflow.com/questions/54135313/webview-extension-in-typescript `VS Code does not include TypeScript typings for the VS Code api available to scripts inside of webviews` we are using vscode not for types but for calling `vscode.postMessage(message);` in extensionIntermediary.

package.json
```diff
@@ -97,20 +107,26 @@
         "chai": "^4.2.0",
         "chalk": "^4.0.0",
         "changelog-parser": "^2.8.0",
+        "chokidar": "^3.5.1",
```
TwitchBronBron, line 110: So, you're gonna laugh at my flip-flip on this "issue". but with the way you have the extension coded, it will fail without chokidar being a dependency, since the import happens at the top of the file. We can negate that by changing the code to use chokidar like this inside your `if` block:
```
if (isDebugMode) {
    const chokidar = require('chokidar');
    ...other logic
}
```

But it's probably just easier to switch this back to a dependency. :D

PR #357, opened 2022-02-04
Fix colorization with interfaces inside namespaces

TwitchBronBron:
Fixes bug where interfaces weren't colored properly when indented.

Before:
image

After:
image


chrisdp (approved):

livecano (commented):
Looks good!

PR #358, opened 2022-02-04
Merge primitive literal exprs into single lib pattern

TwitchBronBron:
Combines the various primitive literal expressions into a single pattern since they never need to be referenced individually. 

This is just a housekeeping PR, no new functionality.


chrisdp (approved):

livecano (commented):
👍

PR #363, opened 2022-03-02
Webpack build

DrunkinMaster:


TwitchBronBron:
Thanks for this, but I'd prefer to use esbuild instead since it's significantly faster build times. I've already got an open PR to handle this, but I haven't tested it in a while and needed to verify that the extensoin host debugging experience was working properly. would you be open to giving that branch a try instead? #335

PR #371, opened 2022-03-19
Better function tmscope handling

### LINKED ISSUE #364: Syntax highlighting issue when using "Function" as an argument type
When using `as Function`, the next two lexemes will display in red (as if it was part of the "function" token) and purple (as if it was a function name).

image

If I add whitespace after `as Function`, then it'll look right.

image

TwitchBronBron:
Refactors the entire function syntax highlighting approach. Followed the patterns from the [typescript grammar](https://github.com/microsoft/vscode/blob/main/extensions/typescript-basics/syntaxes/TypeScript.tmLanguage.json), and matches the highlighting style from typescript and C#. 

**BrightScript:**
image

**TypeScript:**
image

Fixes #364.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,7 @@
                 "@typescript-eslint/eslint-plugin": "^5.14.0",
                 "@typescript-eslint/parser": "^5.14.0",
                 "chai": "^4.2.0",
-                "chalk": "^4.0.0",
+                "chalk": "^4.1.2",
                 "changelog-parser": "^2.8.0",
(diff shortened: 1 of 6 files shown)


TwitchBronBron:
this isn't quite right. it broke anon functions. will need to investigate

image

TwitchBronBron:
Decided to refactor all of the function syntax highlighting logic. Updated PR title and description.

TwitchBronBron:
TODO: class field types need to have same color as param types. 
TODO: class method matching.

Issue #367, opened 2022-03-17
Rendezvous timing issue


TwitchBronBron:
When rendezvous logging is enabled, sometimes entries are missing the completion time. in our rendezvous tree view, right now we're ignoring rendezvous entries if their total time is 0. However, we need to show the total time even if 0ms, because there might be entries with no recorded time where we still want to show the rendezvous count. 

03-17 14:36:29.108 [beacon.signal] |AppLaunchComplete ---------> Pending Render Pass
03-17 14:36:29.109 [sg.node.BLOCK] Rendezvous[2112] at visionsdk:/components/VisionConfig.brs(6)
03-17 14:36:29.638 [sg.node.UNBLOCK] Rendezvous[2112] completed in 0.529 s
03-17 14:36:29.639 [sg.node.BLOCK] Rendezvous[2114] at visionsdk:/components/VisionConfig.brs(6)
03-17 14:36:29.640 [sg.node.UNBLOCK] Rendezvous[2114] completed
03-17 14:36:29.642 [sg.node.BLOCK] Rendezvous[2116] at visionsdk:/components/VisionConfig.brs(15)
03-17 14:36:29.642 [sg.node.UNBLOCK] Rendezvous[2116] completed in 0.001 s
03-17 14:36:29.644 [sg.node.BLOCK] Rendezvous[2119] at visionsdk:/components/VisionConfig.brs(15)
03-17 14:36:29.644 [sg.node.UNBLOCK] Rendezvous[2119] completed in 0.001 s
03-17 14:36:29.881 [sg.node.BLOCK] Rendezvous[2120] at visionsdk:/components/VisionConfig.brs(28)
03-17 14:36:29.881 [sg.node.UNBLOCK] Rendezvous[2120] completed in 0.001 s
03-17 14:36:29.882 [sg.node.BLOCK] Rendezvous[2121] at visionsdk:/components/VisionConfig.brs(33)
03-17 14:36:29.882 [sg.node.UNBLOCK] Rendezvous[2121] completed

Notice the second VisionConfig.brs(6) rendezvous entry is missing a completed time?

Issue #188, opened 2019-07-31
Improvements to how we handle Roku remote inputs


chrisdp:
I am opening this issue to talk over ways we might be able to improve how we send remote events to a Roku as the current approach gets triggered by mistake a lot and is not very user intuitive. One thing we would not want to lose is the easy ability to still view logs while using the remote inputs.

A few brain storm ideas below but would love to hear more ideas!

- This one would need exploring but one way would be a hot key that opens something like a input dialog and we get keys while that’s open but I’m not sure if it's possible to get the key presses in this way. 
- If we get to the point where we have our own side tab we could maybe have a section in that area that, when focused, works the same way as the output does now. 
- Lastly we could possibly do a custom web view that could even allow for some key event automation. 

I would envision all the above could be mapped to a hot key/extension command.

PR #375, opened 2022-03-27
Ninja updates

chrisdp:
Please pay close attention to the `package.json` file. A huge amount of the changes to the remote are based in this area.

Functional changes in this PR:

- Overhaul of when keyboard presses will be send to the device for remote control emulation #188 
  - 
  - The when to send commands condition is no longer tied to being focused in the debug panel or output panel
  - There is now a dedicated `extension.brightscript.toggleRemoteControlMode` that will toggle remote control emulation on and off. You can almost think of it like how you toggle in and out of insert mode in VIM.
    - Also added incase you wish to use use them for keybindings:
     - `extension.brightscript.enableRemoteControlMode`
     - `extension.brightscript.disableRemoteControlMode`
  - When `remoteControlMode` is on all relevant keyboard actions will be sent to the device
    - Exceptions to this rule are when the command pallet, quick pick, inout box, or an input has focus/is open
    - There is a new status bar item that you can click to disable and enable this mode
  - Remote emulation now supports **WAY MORE**!!!
    - for example: a-z, A-Z, 0-9, and all the primary character such as !, @, #, ', ", etc...

- Send text to device now supports command history!
  - 
  - with this change the extension will now remember the past 30 unique strings sent to the device
  - new user settings
    - `brightscript.sendRemoteTextHistory.limit`: Sets the maximum number of 'extension.brightscript.sendRemoteText' commands to remember for future use
      - This is an application setting and is only available at the user settings 
    - `brightscript.sendRemoteTextHistory.enabled`: If true, the extension will not save the history for 'extension.brightscript.sendRemoteText' commands
      - This is a resource settings and is available in the application, machine, window, resource, and machine-overridable settings
  - You will now be presented a history of commands in addition to the input box.
image

- Custom view within the activity bar
  - 
  - We now have our own view in the activity panel
    - this view currently shows you all the devices we have found on the network
    - you can expand the item to see details about the device
    - click on a detail item will copy the value to your clipboard
    - There is also an item that lets you easily open the web portal for the device
image
(image is from prototyping and is missing the open web portal item)

- New and missing ECP commands
  - 
  - `extension.brightscript.pressInstantReplayButton`
  - `extension.brightscript.pressSearchButton`
  - `extension.brightscript.pressEnterButton`
  - `extension.brightscript.pressFindRemote`
  - `extension.brightscript.pressVolumeDown`
  - `extension.brightscript.pressVolumeMute`
  - `extension.brightscript.pressVolumeUp`
  - `extension.brightscript.pressPowerOff`
  - `extension.brightscript.pressChannelUp`
  - `extension.brightscript.pressChannelDown`
  - `extension.brightscript.changeTvInput`

- Bug fixes
  - 
  - Fixed an issue where the rendezvous ui commands would show up in the when they shouldn't be
  - Rendezvous timing issue #367


src/VscodeContext.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type ContextValue = boolean | string;
+
+/**
+ * Wrapper around VS Code's `setContext`.
+ * The API call takes several milliseconds to seconds to complete,
+ * so let's cache the values and only call the API when necessary.
+ */
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+export abstract class VSCodeContext {
```
chrisdp, line 11: Our rule disallows this. I brought this example from [vscodeContext}(https://github.com/VSCodeVim/Vim/blob/a4656a6e53a81e1ae4e2406f81ad831afffd0a18/src/util/vscodeContext.ts) in the vim extension. Found it to be a clean and easy way to mange custom contexts.
Let me know if we should change this.

src/BrightScriptCommands.ts
```diff
@@ -188,6 +252,14 @@ export class BrightScriptCommands {
         }
     }
 
+    private registerCommand(name: string, callback: (...args: any[]) => any, thisArg?: any) {
```
chrisdp, line 255: Just broke it out to clean up the main register function as it was getting a bit noisy.

src/BrightScriptCommands.ts
```diff
@@ -188,6 +252,14 @@ export class BrightScriptCommands {
...
+        this.context.subscriptions.push(vscode.commands.registerCommand('extension.brightscript.' + name, callback, thisArg));
+    }
+
+    private async sendAsciiToDevice(character: string) {
```
chrisdp, line 259: Broke it out for reuse for the keyboard text commands.

src/GlobalStateManager.ts
```diff
@@ -1,13 +1,26 @@
-import type * as vscode from 'vscode';
+import * as vscode from 'vscode';
+
 export class GlobalStateManager {
     constructor(
         private context: vscode.ExtensionContext
-    ) { }
+    ) {
+        this.updateFromVsCodeConfiguration();
+        vscode.workspace.onDidChangeConfiguration(this.updateFromVsCodeConfiguration);
```
chrisdp, line 8: This should work right? I've noticed in a bunch of spots we actually duplicate it all it seems to me most of the time we could have one touch point for reading the first time and when changes happen.

src/BrightScriptCommands.ts
```diff
@@ -60,50 +68,106 @@ export class BrightScriptCommands {
...
+        // Add any that where not cleanly in the range above such as TAB
+        asciiList.push(...['\t']);
+
+        const AsyncFunction = Object.getPrototypeOf(async () => {}).constructor;
+        for (let character of asciiList) {
+            // Escape if needed
+            let escapeCharacter = (character === '\\' || character === '\'') ? '\\' : '';
+
+            // Create the callback function with the correct ascii value in the function body to be called later
+            const dynamicFunction = new AsyncFunction('', `await this.sendAsciiToDevice('${escapeCharacter}${character}');`).bind(this);
```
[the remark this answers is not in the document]
chrisdp: Rewrote this logic to be MUCH cleaner. Resolving this conversation.

TwitchBronBron:
Let's add a "turn remote control on" command and a "turn remote control off" command.

TwitchBronBron:
Let's add a "turn remote control on" command and a "turn remote control off" command.

TwitchBronBron:
Let's add a "turn remote control on" command and a "turn remote control off" command.

TwitchBronBron:
need to allow typing in the debug console search box when remote mode is enabled.

image

TwitchBronBron:
I just added device icons for stick, tv, and set top box. I also added an svg icon for the `BRS` sidebar button.

image
image

src/VscodeContext.ts
TwitchBronBron: Fixed this in a recent commit

TwitchBronBron:
Made the sidebar logo even prettier, and made the lines for set-top-box thicker to match brightness of other logos.
image

TwitchBronBron (approved):

PR #389, opened 2022-05-11
Add component statement colorization

TwitchBronBron:
Adds colorization to component statements

image


chrisdp:
@TwitchBronBron are we doing any of the template colors in this or it will just be considered a template string in the short term?

TwitchBronBron:
No template colors right now. They'll just all show up as the string color. We can get to those once we land the core functionality

PR #401, opened 2022-06-01
Allow RDB to work with user specified port

triwav:


src/extension.spec.ts
```diff
@@ -94,15 +94,13 @@ describe('extension', () => {
 
     describe('RDB', () => {
         const extensionInstance = extension.extension as any;
-        let config;
+        const config = {} as any;
```
TwitchBronBron, line 97: Can we give this a better type? It's probably `BrightScriptLaunchConfiguration` right?
```suggestion
        const config = {} as BrightScriptLaunchConfiguration;
```

src/extension.spec.ts
```diff
@@ -135,19 +133,32 @@ describe('extension', () => {
...
+        describe('debugSessionCustomEventHandler', () => {
+            describe('BSChannelPublishedEvent', () => {
+                const e = {
+                    event: 'BSChannelPublishedEvent',
+                    body: {
+                        launchConfiguration: config
+                    }
+                };
+
+                afterEach(() => {
```
TwitchBronBron, line 145: Is this necessary? We're already doing the exact same thing on line 44. 

https://github.com/rokucommunity/vscode-brightscript-language/pull/401/files#diff-0d79d039c81b1141d7761f91b299b6eb9e7d71f8341fb305652f3bd475f77221R43

src/extension.ts
```diff
@@ -277,11 +288,11 @@ export class Extension {
...
             OnDeviceComponent: {
-                // uncomment for debugging
-                // logLevel: 'verbose',
-                // serverDebugLogging: true,
+                logLevel: enableDebugging ? 'verbose' : undefined,
+                serverDebugLogging: enableDebugging,
                 disableTelnet: true,
-                disableCallOriginationLine: true
+                disableCallOriginationLine: true,
+                callbackListenPort: config.rdbCallbackPort
```
TwitchBronBron, line 295: Just checking, is it ok that this might be undefined sometimes? I don't see any default being set in the config provider. I'm assuming rta will fall back to a random port in that situation?

TwitchBronBron (changes requested):

src/extension.spec.ts
triwav: Removed I was trying to debug why the test was failing and accidentally left in

src/extension.spec.ts
triwav: updated

src/extension.ts
triwav: correct. If it's undefined in the config then it will use the auto find port logic

TwitchBronBron (approved):

TwitchBronBron (approved):

PR #410, opened 2022-07-08
Rdb improvements

triwav:
See commit messages for a list of what all what was done as a part of this PR

https://user-images.githubusercontent.com/1753881/177899012-e9fbad03-d9d2-469a-8614-783f32b1cd16.mov

.

### DIFF
--- a/ui/rdb/components/SceneGraphInspectorPanel/NodeBranchView.svelte
+++ b/ui/rdb/components/SceneGraphInspectorPanel/NodeBranchView.svelte
@@ -14,6 +14,7 @@
     let self: HTMLDivElement;
     let nodeInfoElement: HTMLDivElement;
     let nodeFieldsCache: ODC.NodeRepresentation?;
+    let showNodeInfo = false;
 
     const expandedStorageKey = `expanded:${nodeTree.ref}`;
     export let expanded = utils.getStorageBooleanValue(expandedStorageKey);
@@ -59,7 +60,7 @@
     function startNodeInfoTimeout() {
         stopNodeInfoTimeout();
         nodeInfoHideTimeout = setTimeout(async () => {
-            nodeInfoElement.classList.add('hide');
+            showNodeInfo = false
         }, 100);
     }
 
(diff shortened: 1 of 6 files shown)


ui/rdb/components/SceneGraphInspectorPanel/NodeBranchView.svelte
```diff
@@ -51,10 +55,125 @@
         dispatch('open', nodeTree);
     }
 
+    let nodeInfoHideTimeout: ReturnType<typeof setTimeout>;
+    function startNodeInfoTimeout() {
+        stopNodeInfoTimeout();
+        nodeInfoHideTimeout = setTimeout(async () => {
+            nodeInfoElement.classList.add('hide');
```
TwitchBronBron, line 62: I think we ran into issues doing manual css class modifications in the past with svelte. Shouldn't this really be toggling a `.visible` property on the data, and let svelte show/hide it?

ui/rdb/components/SceneGraphInspectorPanel/NodeDetailView.svelte
```diff
@@ -130,6 +126,62 @@
...
+    }
+
+    function onResetValueButtonClicked() {
+        const elements = [];
+        for (const child of this.parentElement.children) {
+            if (child.classList.contains('fieldValue')) {
+                elements.push(child);
+            };
+        }
+        if (elements.length === 1) {
```
TwitchBronBron, line 164: Please add comments about WHY the length is 1 or 2. 

What happens if length is 0 or > 2? Should those be handled?

TwitchBronBron (changes requested):
Two small things, but overall it looks great!

ui/rdb/components/SceneGraphInspectorPanel/NodeBranchView.svelte
triwav: Updated here. We do use it in the NodeDetailView and it seems to work ok but this one is a simple 1:1 relationship so I switched it

ui/rdb/components/SceneGraphInspectorPanel/NodeDetailView.svelte
triwav: Added comments and a console log in case we ever hit it

TwitchBronBron (approved):

PR #422, opened 2022-08-12
Add .bs and .brs file icon support

TwitchBronBron:
Adds custom file icons for .bs and .brs files. 

Note that the brightscript icon is slightly wider than most other icons. I thought it was important to show the lowercase `r` to better visualize that this was a non-brighterscript file. 

image


TwitchBronBron:
Another option to prevent the larger file icon would be to draw the `brs` in all lower case like this:
image

TwitchBronBron:
Went with the second option (lower case `brs`)

PR #455, opened 2022-12-27
Fix rdb keypaths and focus scrolling

triwav:
• Fix focus scrolling to work properly the first time even if all levels were collapsed
• update to RTA v2.0 beta 11
• Update absolute key path to properly be able to dig into ArrayGrids and use `#` before findNode keys
• expand focused node for easier access to its children


webviews/src/views/SceneGraphInspectorView/NodeDetailPage.svelte
```diff
@@ -17,8 +17,24 @@
         let nodeTree = inspectNodeNodeTree;
         if (inspectNodeNodeTree) {
             while (nodeTree) {
-                if (nodeTree.id) {
-                    absoluteKeyPathParts.unshift(nodeTree.id);
+                if (nodeTree.subtype === 'RowListItem') {
```
TwitchBronBron, line 20: I don't fully understand why we're modifying the keypath structure here. Shouldn't we keep the full keypath as RTA provides?

PR #472, opened 2023-03-23
Added validation for raleTrackerTaskFileLocation setting existence.

jean-guyrivard:
Ran into this issue on a new install of an existing project.
injectRale was set, setting was non-existent and error VS Code gave was...less than helpful.


TwitchBronBron (changes requested):
Thanks for your contribution! I just had one small suggestion on the wording of the error, and then I'm happy to merge and release this.

jean-guyrivard:
Approved change request.
Either text was fine with me as they are miles better than the error VS Code threw.

PR #478, opened 2023-04-26
Fix wsl rdb bugs

TwitchBronBron:
- Fixes crash in the rdb webview that wouldn't work in WSL.
- refines the setup instructions when no odc was found


webviews/src/shared/OdcSetupSteps.svelte
```diff
@@ -7,18 +7,28 @@
...
+            Enable the following property in <b><i>launch.json</i></b>:
+            <pre><code>"injectRdbOnDeviceComponent": true</code></pre>
+        </li>
+        <li>
+            Add this comment to <code>function main()</code> after the call to <code>screen.show()</code>
+            <pre><code>' vscode_rdb_on_device_component_entry</code></pre>
+        </li>
+    </ol>
+
+    The ODC is properly installed if you see this log output during channel launch:
```
triwav, line 31: The ODC component?

TwitchBronBron: Odc means on device component. So you want it to say "on-device-component component"? 😜

I'd rather spell it out in that case.

triwav: Maybe `The on-device component is properly installed if you see this log output during channel launch:`

PR #496, opened 2023-08-07
Add roku automation panel

triwav:
Add a new panel to the extension that allows running a series of commands that can be automatically run on deploy or run manually.


triwav:
Closing as combining with new detail view work

PR #503, opened 2023-09-26
Enable remote control on launch

### LINKED ISSUE #475: Add launch setting to manage Remote Control Mode
Add a setting to `launch.json` that allows for controlling the "Remote Control Mode" whenever the debug session launches and ends. 
The primary options I can think of right now are:
 -  "turn off when session ends"
 - "turn on when session starts, turn off when session ends"

iBicha:
Closes #475

### DIFF
--- a/package.json
+++ b/package.json
@@ -825,10 +825,27 @@
                                 "default": false,
                                 "description": "Delete any currently installed dev channel before starting the debug session"
                             },
-                            "enableRemoteControl": {
-                                "type": "boolean",
-                                "default": true,
-                                "description": "Enable remote control mode when starting the debug session"
+                            "remoteControlMode": {
+                                "oneOf": [
+                                    {
+                                        "type": "object",
+                                        "description": "Options for activating and deactivating remote control mode",
+                                        "properties": {
+                                            "activateOnSessionStart": {
+                                                "type": "boolean",
+                                                "description": "Activate remote control mode on debug session start"
+                                            },
+                                            "deactivateOnSessionEnd": {
+                                                "type": "boolean",
+                                                "description": "Deactivate remote control mode on session end"
+                                            }
+                                        }
+                                    },
+                                    {
+                                        "type": "boolean",
+                                        "description": "Activate on session start, deactivate on session end."
+                                    }
+                                ]
                             }
                         }
                     }
--- a/package.json
+++ b/package.json
@@ -824,6 +824,11 @@
                                 "type": "boolean",
                                 "default": false,
                                 "description": "Delete any currently installed dev channel before starting the debug session"
+                            },
+                            "enableRemoteControl": {
+                                "type": "boolean",
+                                "default": true,
+                                "description": "Enable remote control mode when starting the debug session"
                             }
                         }
                     }
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -52,7 +52,8 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
             enableDebugProtocol: false,
             remotePort: 8060,
             rendezvousTracking: true,
-            deleteDevChannelBeforeInstall: false
+            deleteDevChannelBeforeInstall: false,
+            enableRemoteControl: false
         };
 
         let config: any = vscode.workspace.getConfiguration('brightscript') || {};
@@ -243,6 +244,7 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
         config.cwd = folderUri.fsPath;
         config.rendezvousTracking = config.rendezvousTracking === false ? false : true;
         config.deleteDevChannelBeforeInstall = config.deleteDevChannelBeforeInstall === true;
+        config.enableRemoteControl = config.enableRemoteControl === true ? true : this.configDefaults.enableRemoteControl;
 
         if (config.request !== 'launch') {
             await vscode.window.showErrorMessage(`roku-debug only supports the 'launch' request type`);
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -824,6 +824,11 @@
                                 "type": "boolean",
                                 "default": false,
                                 "description": "Delete any currently installed dev channel before starting the debug session"
+                            },
+                            "enableRemoteControl": {
+                                "type": "boolean",
```
TwitchBronBron, line 829: Let's expand this _slightly_ to accomodate some of the options described in the ticket.
image

```
 "remoteControlMode": {
    "oneOf": [
        {
            "type": "object",
            "description": "Options for activating and deactivating remote control mode",
            "properties": {
                "activateOnSessionStart": {
                    "type": "boolean",
                    "description": "Activate remote control mode on debug session start"
                },
                "deactivateOnSessionEnd": {
                    "type": "boolean",
                    "description": "Deactivate remote control mode on session end"
                }
            }
        },
        {
            "type": "boolean",
            "description": "Activate on session start, deactivate on session end."
        }
    ]
}
```

src/DebugConfigurationProvider.ts
```diff
@@ -561,4 +563,10 @@ export interface BrightScriptLaunchConfiguration extends LaunchConfiguration {
      * If injectRdbOnDeviceComponent is true and this is true the screen saver will be be disabled while the deployed application is running.
      */
     disableScreenSaver?: boolean;
+
+    /**
+     * If true, the remote control will be enabled at the start of the debug session
+     * @default false
+     */
+    enableRemoteControl?: boolean;
```
TwitchBronBron, line 571:
```suggestion
    remoteControlMode?: {
        activateOnSessionStart: boolean;
        deactivateOnSessionEnd: boolean;
    };
```

iBicha:
> This is great! I have a few suggestions on the general design of the launch.json settings.
> 
> And can you add a unit test or two ensuring that the `"remoteControlMode": true` turns into `"remoteControlMode": {...the stuff}`

Thanks for the review - I think I got what you wanted out of this feature (I personally prefer to hit Command+K for the remote, this is why I haven't considered the different possible preferences)

src/DebugConfigurationProvider.ts
```diff
@@ -561,4 +567,10 @@ export interface BrightScriptLaunchConfiguration extends LaunchConfiguration {
      * If injectRdbOnDeviceComponent is true and this is true the screen saver will be be disabled while the deployed application is running.
      */
     disableScreenSaver?: boolean;
+
+    /**
+     * If true, the remote control will be enabled at the start of the debug session, and disabled at the end of the debug session.
+     * @default false
+     */
+    remoteControlMode?: boolean | { activateOnSessionStart?: boolean; deactivateOnSessionEnd?: boolean };
```
TwitchBronBron, line 575: Let's exclude the `boolean` part. The boolean shorthand can be converted to this in our DebugConfigurationProvider.

src/extension.ts
```diff
@@ -177,6 +177,10 @@ export class Extension {
             //if this is a brightscript debug session
             if (e.type === 'brightscript') {
                 this.chanperfStatusBar.hide();
+                const config = e.configuration as BrightScriptLaunchConfiguration;
+                if (config.remoteControlMode && (config.remoteControlMode === true || config.remoteControlMode?.deactivateOnSessionEnd === true)) {
```
TwitchBronBron, line 181: Just check the `.deactivateOnSessionEnd`.
Probably something like `config?.remoteControlMode?.deactivateOnSessionEnd=== true` since our `DebugConfigurationProvider` will have sanitized this to always be in this structure.

src/DebugConfigurationProvider.ts
```diff
@@ -243,6 +247,8 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
         config.cwd = folderUri.fsPath;
         config.rendezvousTracking = config.rendezvousTracking === false ? false : true;
         config.deleteDevChannelBeforeInstall = config.deleteDevChannelBeforeInstall === true;
+        config.remoteControlMode = config.remoteControlMode ? config.remoteControlMode : this.configDefaults.remoteControlMode;
```
TwitchBronBron, line 250: Let's convert this exclusively to the object here, so downstream we can always just check the `activateOnSessionStart`/`deactivateOnSessionEnd` flags. Something more like this?

```typescript

if (typeof config.remoteControlMode === 'boolean') {
    config.remoteControlMode = {
        activateOnSessionStart: config.remoteControlMode,
        deactivateOnSessionEnd: config.remoteControlMode,
    }
} else {
   config.remoteControlMode = {
        activateOnSessionStart: config.remoteControlMode?.activateOnSessionStart ??  this.configDefaults.remoteControlMode.activateOnSessionStart,
        deactivateOnSessionEnd: config.remoteControlMode?.deactivateOnSessionEnd??  this.configDefaults.remoteControlMode.deactivateOnSessionEnd,
    };
}

```

src/extension.ts
```diff
@@ -212,6 +216,9 @@ export class Extension {
             const config = e.body as BrightScriptLaunchConfiguration;
             await docLinkProvider.setLaunchConfig(config);
             logOutputManager.setLaunchConfig(config);
+            if (config.remoteControlMode && (config.remoteControlMode === true || config.remoteControlMode?.activateOnSessionStart === true)) {
```
TwitchBronBron, line 219: Just check the `.activateOnSessionStart`.
Probably something like `config?.remoteControlMode?.activateOnSessionStart === true` since our DebugConfigurationProvider will have sanitized this to always be in this structure.

TwitchBronBron (changes requested):
Hey! Much closer! I think it would be nicer to eliminate the `boolean` in the interface, so that it's just a shorthand in the config and we can more consistently interact with the activate/deactivate props downstream.

iBicha:
> Hey! Much closer! I think it would be nicer to eliminate the `boolean` in the interface, so that it's just a shorthand in the config and we can more consistently interact with the activate/deactivate props downstream.

That's a good idea, I applied your suggestions

TwitchBronBron (approved):
Everything looks great. Thanks for this!

PR #505, opened 2023-10-09
Add ability to capture device screenshots

fumer-fubotv:
1. Ability to add support for "capture screenshot" through command palette and device view.
2. Upon success - screenshot preview is shown to user

Detailed issue: https://github.com/rokucommunity/vscode-brightscript-language/issues/430

### DIFF
--- a/package.json
+++ b/package.json
@@ -2778,19 +2778,19 @@
             },
             {
                 "command": "extension.brightscript.rokuDeviceView.pauseScreenshotCapture",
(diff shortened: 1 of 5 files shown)


TwitchBronBron (approved):
I made a few tweaks to the password logic flow to make it work better with existing launch.json password, and also fixed a few crashes related to missing passwords. I added an emoji to the tree item to make it stand out more as a command. 

This looks great! Thanks for your great work on this.

PR #514, opened 2023-11-01
Add app dropdown menu for ECP registry link

MilapNaik:
Issue: https://github.com/rokucommunity/vscode-brightscript-language/issues/510

Added a quickpick dropdown menu that shows the apps list on a device (seen below) so a user can pick an app that they are possibly keyed for before opening the ECP registry

image

Here's an example of this in action:
registry-link


src/viewProviders/OnlineDevicesViewProvider.ts
```diff
@@ -49,7 +51,80 @@ export class OnlineDevicesViewProvider implements vscode.TreeDataProvider<vscode
...
+                }
+            });
+
+            client.on('error', (error) => {
+                console.error(`Error: ${error.message}`);
+                client.destroy();
+                reject(error);
+            });
+        });
+    }
```
TwitchBronBron, line 90: This seems fairly verbose. We already imported a request library in this project. You could do something like this instead:

in util.ts, add this code:
```typescript
import * as request from 'postman-request';

export class Util {
   // ...
    public httpGet(url: string) {
        return new Promise<Response>((resolve, reject) => {
            request.get(url, (err, response) => {
                return err ? reject(err) : resolve(response);
            });
        });
    }
   ///...
}
```

Then you can just call:

```typescript
const data = await util.httpGet(`${deviceIp}/query/apps`);
```

src/BrightScriptCommands.ts
```diff
@@ -263,6 +263,24 @@ export class BrightScriptCommands {
             }
         });
 
+        this.registerCommand('openQuickInputBox', async (url: string, placeholder: string, appNames: vscode.QuickPickItem[]) => {
```
TwitchBronBron, line 266: This command should have a more meaningful name. 

Also, I think it should just fetch the app data itself rather than needing them to be passed in. So something like this:
```typescript
this.registerCommand('openRegistry', async (host: string) => {
    if (!host) {
        host = await this.promptForHost();
    }
    
     const apps = await util.httpGet(`${host}:8060/query/apps`);
      //convert the apps to QuickPickItems
      
      //show the quickPick
      
      //open the URL if all is good

TwitchBronBron (changes requested):

src/BrightScriptCommands.ts
TwitchBronBron: Yeah, I think the quick pick is already simple enough when used directly from the vscode APIs, so I think it's more valuable to make this command specific to opening the registry.

src/viewProviders/OnlineDevicesViewProvider.ts
MilapNaik: ok, changes made

src/BrightScriptCommands.ts
MilapNaik: done

TwitchBronBron (approved):

PR #534, opened 2023-12-06
Support a configurable port for SceneGraphDebugCommandController

Christian-Holbrook:


package.json
```diff
@@ -830,6 +830,11 @@
                                 "default": false,
                                 "description": "Delete any currently installed dev channel before starting the debug session"
                             },
+                            "sceneGraphDebugCommandsPort": {
+                                "type": "number",
+                                "default": 8080,
+                                "description": "Delete any currently installed dev channel before starting the debug session"
```
TwitchBronBron, line 836: I don't think this description is right...

PR #536, opened 2023-12-20
Always request screenshot on initial load, add ability to copy screenshot

triwav:
Some small improvements to screenshot capability to make it easier to share screenshots from device. Changed it to request an initial screenshot where as before if you didn't have continuous screenshot capture on it would not request any on its own unless you clicked refresh

### DIFF
--- a/package.json
+++ b/package.json
@@ -302,6 +302,11 @@
                     "group": "navigation@3"
                 },
                 {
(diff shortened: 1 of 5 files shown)


TwitchBronBron (approved):
Looks pretty straightfoward. Nice!

PR #537, opened 2024-01-08
Pass the modal property from the PopupMessageEvent

Christian-Holbrook:
This change utilizes the modal property that was added to the `PopupMessageEvent`. The modal property is being added to make any errors more visible when VSCode fails to launch a channel. The most common example of this is a bad password.

Here's a screen shot of the modal.
Screenshot 2024-01-08 091210


TwitchBronBron:
@Christian-Holbrook can you install the very latest version of roku-debug? I _just_ released the change with that new `modal` option, which is why the CI is failing.

PR #548, opened 2024-02-29
Device disconnect support and other improvements

triwav:
add support for disconnecting from device in webview panels, avoid issues with deploying due to screenshot capture happening at same time, add support for copying node info response in scenegraph inspector detail view


TwitchBronBron (commented):
Everything looks good. One suggestion would be to re-enable the device view if the debug session crashes for some reason. There might be a "vscode.debug.onDidStopDebugSession" event? Not a blocker though.

TwitchBronBron (approved):

PR #552, opened 2024-03-10
Delete button in FS view

iBicha:
fs-delete

### DIFF
--- a/webviews/src/views/RokuFileSystemView/FileSystemEntry.svelte
+++ b/webviews/src/views/RokuFileSystemView/FileSystemEntry.svelte
@@ -1,6 +1,6 @@
 
 <script lang="ts">
-    import { File, Database, Folder } from 'svelte-codicons';
+    import { File, Database, Folder, Trash } from 'svelte-codicons';
     import { createEventDispatcher } from 'svelte';
     import type { PathContentsInfo } from '../../shared/types';
     const dispatch = createEventDispatcher();
@@ -15,6 +15,10 @@
 
     function onDoubleClick() {
         dispatch('open', entry);
+    }
+
+    function onDeleteClick() {
+        dispatch('delete', entry);
     }
 
     function padNum(number: number) {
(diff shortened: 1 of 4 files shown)


webviews/src/views/RokuFileSystemView/FileSystemEntry.svelte
```diff
@@ -17,6 +17,10 @@
         dispatch('open', entry);
     }
 
+    function onDeleteClick() {
+        dispatch('delete', entry);
```
TwitchBronBron, line 21: I'm thinking we should add a confirmation dialog before executing the delete? Might need to do it in the vscode side?

iBicha: I personally wouldn't just for how I'm using it:
- I'm deleting 10-20 files from cachefs, to get my app to recache things. I'm also not deleting everything, so a button to delete all of cachefs is not the answer (although it could be a good addition in a future PR - I didn't add that yet since there was no equivalent for it in ODC). A pop-up per file would probably irritate me and I wouldn't use this
- You'd mostly use this to delete tmp or cachefs, which is by no means critical if deleted accidentally. I guess there's the case for additional volumes, but that's way less common

chrisdp: I mostly agree that it's likely not mission critical. @TwitchBronBron @iBicha maybe a compromise is a setting to turn off the confirmation?

iBicha:
A weird thing is that files under `common:` return permissions `rw`. So the delete button is there but won't work because we don't have permissions to delete the files. Perhaps this kind of filtering should be based on the returned permissions + hardcoded exceptions (like `common:`). But it's also a very minor thing, I don't think anyone would be confused why `common:` files are not being deleted.

webviews/src/views/RokuFileSystemView/FileSystemEntry.svelte
TwitchBronBron: Okay. If it's not worth it, no need adding complexity.

iBicha: Let me know what's a nice way to proceed
- No pop-up
- Pop-up each time
- Pop-up with a note that there's a vscode setting to disable it

TwitchBronBron: We can ship without a popup. If someone wants that, they can ask or implement it. :)

triwav (approved):

PR #561, opened 2024-04-09
Add support for opening SceneGraph Inspector in panel

triwav:
This PR allows opening the SceneGraph Inspector as a WebViewPanel in the current window. As of the April 2024 release of Visual Studio Code, you can also move that panel from the current window to a new window as well although there is currently no way to programmatically do this.
Apr-09-2024 12-47-46


TwitchBronBron:
I made one small tweak, to use `link-external` icon instead of `empty-window`. 

image

TwitchBronBron (approved):

PR #563, opened 2024-04-17
Add support for focusing a node in SceneGraph Inspector

triwav:
Apr-17-2024 15-23-13


webviews/src/views/SceneGraphInspectorView/Branch.svelte
```diff
@@ -265,6 +271,12 @@
         </span>
     </div>
     <div class="actions">
+        {#if treeNode.visible !== undefined}
```
TwitchBronBron, line 274: What if `tree` is `false`? Shouldn't this be:
```
{#if treeNode.visible === true}
```

TwitchBronBron: nevermind, i get it now.

PR #576, opened 2024-06-04
Fix lsp crash tracker

TwitchBronBron:
Fix a bug in the lsp crash tracker.

### DIFF
--- a/src/LanguageServerManager.ts
+++ b/src/LanguageServerManager.ts
@@ -44,9 +44,10 @@ class LspRunTracker {
     }
 
     public setState(state: State) {
-        clearTimeout(this.timeoutHandle);
         //if language server is running, clear any timers
-        if (state === State.Stopped) {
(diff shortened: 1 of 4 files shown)


src/LanguageServerManager.ts
```diff
@@ -168,75 +223,34 @@ export class LanguageServerManager {
...
-            this.client = new LanguageClient(
-                'brighterScriptLanguageServer',
-                LANGUAGE_SERVER_NAME,
-                serverOptions,
-                clientOptions
-            );
-            // Start the client. This will also launch the server
-            this.clientDispose = this.client.start();
-            await this.client.onReady();
             this.client.onDidChangeState((event: StateChangeEvent) => {
```
TwitchBronBron, line 228: This was the primary fix. We need to subscribe to these events _before_ starting the client so we pick up the `starting` and `running` states.

PR #582, opened 2024-07-25
Sort imports setting

iBicha:
I saw this option missing - is this all it takes to make it available?


TwitchBronBron (approved):
Yeah, I think this is all that was missing.

PR #583, opened 2024-07-30
Adds ability to install a missing bsc version

TwitchBronBron:
Adds the ability to install brighterscript versions automatically without requiring it to exist in `${workspaceFolder}/node_modules`.

1. After this PR, the bsdk value can support the following values:

    - an absolute path to a preinstalled version of brighterscript (i.e. `C:/roku/brighterscript`)
    - a relative path to a preinstalled version of brighterscript (i.e. `./node_modules/brighterscript`)
    - an exact version of brighterscript from the npm registry (i.e. `0.67.6`, `1.0.0-alpha.36`)
    - an absolute path to a `.tgz` file (i.e. `C:/downloads/brighterscript-0.67.6.tgz`)
    - a relative path to a `.tgz` file (i.e. `./packages/brighterscript-0.67.6.tgz`)
    - a URL to a .tar file (i.e. `npm install https://github.com/rokucommunity/brighterscript/releases/download/v0.67.6/brighterscript-0.67.6.tgzz`)

2. We've added a new setting called `npmCacheRetentionDays` which determines how long to wait before deleting unused brighterscript versions from the local cache. 
3. Added a new command called `clearNpmPackageCache` which will clear all downloaded brighterscript packages
4. Added a new series of menu options for downloading brighterscript versions, and also a few maintenance items like opening the install directory or clearing the downloaded files
    image

Here's a demo of the feature in action:

https://github.com/user-attachments/assets/95b09ad5-a5a0-46f3-b960-81f296d962db

TODO:

- [x] add menu item/command to clear all cached versions
- [x] delete bsc versions that are older than 45 days (make this configurable)
- [x] add `vscode.window.onDidChangeWindowState` event to mark a bsc version's "last used date"
- [x] show all npm versions in the dropdown
- [x] show mainline and prerelease versions in separate menus
- [x] verify the extension storage gets cleaned up on extension uninstall (there's no way to do this, it's probably fine?)


package.json
```diff
@@ -3084,6 +3085,12 @@
                 "title": "Open SceneGraph Inspector In New Window",
                 "category": "BrighterScript",
                 "icon": "$(link-external)"
+            },
+            {
+                "command": "extension.brightscript.clearNpmPackageCache",
+                "title": "Clear the extension's local node_modules cache",
+                "category": "BrighterScript",
+                "icon": "$(link-external)"
```
TwitchBronBron, line 3093: @chrisdp what do you think about this command name and the message?

src/commands/LanguageServerInfoCommand.ts
```diff
@@ -21,6 +28,13 @@ export class LanguageServerInfoCommand {
                 label: `View language server logs`,
                 description: ``,
                 command: this.focusLanguageServerOutputChannel.bind(this)
+            }, {
+                label: `Delete cached brighterscript versions`,
```
TwitchBronBron, line 32: @chrisdp what do you think of this message?

PR #597, opened 2024-09-28
Add feature to be able to set a device as the current active device via the Device tree view. Also added command to clear the active device.

spoyser:
Add feature to be able to set a device as the current active device via the Device tree view.
Also added command to clear the active device.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2731,6 +2731,12 @@
                 "category": "Brightscript"
             },
             {
+                "command": "extension.brightscript.clearActiveDevice",
+                "title": "Clear Active Device",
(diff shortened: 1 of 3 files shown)


spoyser:
I added the ClearActiveDevice command for a common use case (for me anyway) where the device you want to control is not showing in the tree view (e.g. not in developer mode).
By clearing the active device, next time it is needed VSCode will prompt you for it, assuming you have "brightscript.remoteControl.host":"${promptForHost}" in your VSCode settings.

TwitchBronBron (approved):
This is great. Thank you for keeping it so focused. The PR was easy to review, and I think the feature works well. Nice work!

PR #609, opened 2025-01-13
Retain original newlines in BrightScript Log output panel, and also strip ansii colors

chrisdp:
Retain the original newlines in the `BrightScript Log` output panel that were sent from the device (we had some if statements excluding them). Also strip ansii colors from the output panel because they're very ugly and not supported.


src/LogOutputManager.ts
```diff
@@ -258,6 +259,12 @@ export class LogOutputManager {
      */
     public appendLine(lineText: string, mustInclude = false): void {
         let lines = lineText.split(/\r?\n/g);
+
+        // Remove then last line if it's empty as a result of a trailing newline
```
TwitchBronBron, line 263:
```suggestion
        // Remove the last line if it's empty as a result of a trailing newline
```

TwitchBronBron (approved):

PR #611, opened 2025-01-30
issue 197 - incorrect colors used on key words

philanderson888:


philanderson888:
The fix I have generated appears to fix the issue for the given colours.

Please see below.

@TwitchBronBron @chrisdp @tomek-r

Screenshot 2025-01-30 at 17 11 25

philanderson888:
```
Hmm. At first glance, I think this is not the correct approach.

We don't have a lot of unit tests around the syntaxes, so I don't have full confidence as to why these deleted items are still needed, but I'm thinking a better approach would be to add the correct colorizing for those keywords rather than delete whatever else is colliding with them.

Also, if you do end up deleting these, you'll need to remove their references from the list above too, (look for #if_with_paren and #function_call
```

@TwitchBronBron I will take a look again 

thanks

tomek-r:
@philanderson888 How can I test the branch in my VSC?

philanderson888:
when i revert all code just to see the test status on null change, I am getting one failure and the rest passing or skipped

Screenshot 2025-01-31 at 13 29 01

Screenshot 2025-01-31 at 13 29 47

... so at least we know the starting point ... before any code is pushed ....

@TwitchBronBron just for info.

philanderson888:
Thank you for your comments yesterday @TwitchBronBron - that supplied the information that I was missing

After adding this trigger line

```json
                {
                    "include": "#while_with_paren"
                },
```

then yes the 'while' code is doing its job 

so that all works now, and I reverted the existing code so that is all back 'as-it-was' 

thanks

I have been through all the comments and pictures in the issue/pull request and assembled them together in one file - looks like all issues addresssed as part of this issue gathering over the years have now been cumulatively resolved.

Here is what I am seeing for example

Screenshot 2025-01-31 at 13 38 30

... so I have gone through all conversations, all images and attempted to recreate the named items/functions afresh in one file ... so all can be seen together ... all the colours appear to be working correctly now.

@TwitchBronBron @chrisdp @tomek-r

Issue #549, opened 2024-03-05
Add thumbnail to Roku AppOverlays grid


TwitchBronBron:
It would be nice to show a thumbnail in the Roku App Overlays list for each icon. Maybe even make the icon clickable which would open the image in vscode (or show in explorer). 

Before:
![image](https://github.com/rokucommunity/vscode-brightscript-language/assets/2544493/229c89d8-6649-4324-be7c-72eff7777df9)

After:

![image](https://github.com/rokucommunity/vscode-brightscript-language/assets/2544493/fa4aeb17-23c4-4c0f-8435-cd733ac67aa3)

jtuckerfubo:
![Image](https://github.com/user-attachments/assets/dea8c441-c144-40ce-a644-7f0766d4a525)

PR #614, opened 2025-02-06
Add thumbnail to Roku AppOverlays grid #549

jtuckerfubo:
Added a thumbnail for each entry in the app overlay list. Clicking the thumbnail launches an editor view with the source image.

Tested on a Mac using a Roku OS-14/benjamin device


jtuckerfubo:
https://github.com/rokucommunity/vscode-brightscript-language/issues/549

jtuckerfubo:
image

src/viewProviders/BaseWebviewViewProvider.ts
```diff
@@ -263,7 +263,7 @@ export abstract class BaseWebviewViewProvider implements vscode.WebviewViewProvi
             enableScripts: true,
 
             localResourceRoots: [
-                vscode.Uri.file(this.webviewBasePath)
+                vscode.Uri.file('/')
```
TwitchBronBron, line 266: I don't think this change is correct. webviewBasePath is set to `path.join(extensionContext.extensionPath, 'dist', 'webviews');`

src/viewProviders/BaseWebviewViewProvider.ts
```diff
@@ -294,7 +294,7 @@ export abstract class BaseWebviewViewProvider implements vscode.WebviewViewProvi
                     // Enable javascript in the webview
                     enableScripts: true,
                     localResourceRoots: [
-                        vscode.Uri.file(this.webviewBasePath)
+                        vscode.Uri.file('/')
```
TwitchBronBron, line 297: Same with this one, I think this change should be reverted

src/viewProviders/RokuAppOverlaysViewViewProvider.ts
```diff
@@ -27,6 +35,7 @@ export class RokuAppOverlaysViewViewProvider extends BaseRdbViewProvider {
                 }
             };
             const filePath = (await vscode.window.showOpenDialog(options))[0]?.fsPath;
+            const imagePath = path.relative(this.webviewBasePath, filePath);
```
TwitchBronBron, line 38: Should we have any type of fallback image if the path doesn't exist anymore? How does the UI currently handle that?

src/viewProviders/BaseWebviewViewProvider.ts
jtuckerfubo: done

src/viewProviders/RokuAppOverlaysViewViewProvider.ts
jtuckerfubo: When removing and then re-adding images, the UI remains stable (thumbnails are stable, opening images in workspace editor is stable, etc.) ODC becomes unresponsive (until you restart the channel app). I can look into that. I think we technically do have a fallback image (the encoded image data is now preserved on disk).

jtuckerfubo:
image

TwitchBronBron:
That smiley face creeps me out! 🤣

TwitchBronBron:
I made a few tweaks to the layout, utilizing css grid (that thing is awesome!!!). Here's what they look like now. 
image

TwitchBronBron:
And here's the final layout. I'm happy with this, let's ship it!

image

TwitchBronBron (approved):

PR #621, opened 2025-02-19
Adding ability to autogenerate "hello world" template for new developers 

Gchorba:
Adding the ability to autogenerate a "hello world" template for new developers 

Added createNewProject.ts and updated extension.ts and package.json


src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    const folderUri = await vscode.window.showOpenDialog(options);
+    return folderUri?.[0]?.fsPath;
+}
+
+/**
+ * Creates the basic project folder structure
+ */
+async function createProjectStructure(projectPath: string) {
+    const folders = [
```
chrisdp, line 76: We probably want to encourage localization so we should maybe include the locale folder as well.

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+major_version=1
+minor_version=0
+build_version=0
+mm_icon_focus_hd=pkg:/images/channel-poster_hd.png
+mm_icon_focus_sd=pkg:/images/channel-poster_sd.png
+
+splash_screen_sd=pkg:/images/splash-screen_sd.jpg
+splash_screen_hd=pkg:/images/splash-screen_hd.jpg
+
+ui_resolutions=hd
```
chrisdp, line 140:
```suggestion
ui_resolutions=fhd
```

Most developers design in FHD I think.

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+
+
+## Building and Deploying
+
+Use the VS Code BrightScript extension commands to build and deploy your channel:
+
+First, you'll need to configure your launch.json file for debugging. Here's how to set it up:
+
+In VS Code, go to the Run and Debug view (Ctrl+Shift+D or Cmd+Shift+D on Mac)
+Click "create a launch.json file"
```
chrisdp, line 224: I think we could make this as part of the set up if it was missing maybe 🤔

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+    // Create .gitignore
+    const gitignoreContent = `
+out/
+dist/
+.roku-deploy-staging/
+*.zip
+    `.trim();
+
+    // Write all files
+    const files = [
```
chrisdp, line 245: We should include a translations file and maybe even put the translation for the label in there as well.

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+        { path: 'manifest', content: manifestContent },
+        { path: 'source/main.brs', content: mainContent },
+        { path: 'components/MainScene.xml', content: mainSceneContent },
+        { path: 'components/MainScene.brs', content: mainSceneBrsContent },
+        { path: 'README.md', content: readmeContent },
+        { path: '.gitignore', content: gitignoreContent }
+    ];
+
+    for (const file of files) {
+        await fs.promises.writeFile(
```
chrisdp, line 255: I wonder if we should have protections in place before we write the files to make sure we don't replace files by the name name if they already exist.

chrisdp (commented):
Just a small initial pass of review. Need to give it some testing as well but looking good at a glance.

src/commands/createNewProject.ts
Gchorba: yeah, I was going to include a launch.json in the actual sample project but I couldn't figure out how stop it from erroring out with  "rootDir": "${workspaceFolder}" in typescript. I'll go back and add it when I have a moment

src/commands/createNewProject.ts
Gchorba: Yeah, I just don't have a FHD poster image right now. Let me find that

src/commands/createNewProject.ts
Gchorba: What do you mean @chrisdp? can you explain more?

src/commands/createNewProject.ts
Gchorba: fixed this in the new code

src/commands/createNewProject.ts
Gchorba: Done

src/commands/createNewProject.ts
Gchorba: added

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * Creates a new Roku project with a standard directory structure and template files
+ * @param context The extension context, used to access extension resources
+ */
+export async function createNewRokuProject(context: vscode.ExtensionContext) {
+    try {
+        // Get workspace folder or ask user to select a folder
+        const workspaceFolder = await selectWorkspaceFolder();
+        if (!workspaceFolder) {
+            return;
```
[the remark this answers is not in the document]
Gchorba: added

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        // Get project name from user
+        const projectName = await vscode.window.showInputBox({
+            prompt: 'Enter the name for your new Roku project',
+            validateInput: validateProjectName
+        });
+
+        if (!projectName) {
+            return;
```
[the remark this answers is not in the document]
Gchorba: added

src/commands/createNewProject.ts
chrisdp: @Gchorba see this doc about using the `locale` folder for storing string translations. https://developer.roku.com/en-ca/docs/developer-program/core-concepts/localization.md

Happy to chat more if this doesn't clear things up.

Gchorba: Or are you suggesting also doing the images?

Gchorba: @chrisdp checking on this

package.json
```diff
@@ -2639,6 +2652,11 @@
                 "title": "Show Release Notes",
                 "category": "Brightscript"
             },
+            {
+                "command": "extension.brightscript.createNewProject",
```
TwitchBronBron, line 2656: Let's remove the Roku: prefix here, it already shows up as "BrightScript: ` in the command palette, I think that's enough.

src/BrightScriptCommands.ts
```diff
@@ -55,6 +56,13 @@ export class BrightScriptCommands {
             this.activeDeviceManager.refresh();
         });
 
+        // Register the new project creation command
+        this.context.subscriptions.push(
```
TwitchBronBron, line 60: This is doing the same work as down below in line 347, so you can delete this. It actually causes errors in vscode to double-register the same command.

src/commands/createNewProject.ts
```diff
@@ -0,0 +1,352 @@
...
+    await fs.promises.writeFile(
+        launchPath,
+        JSON.stringify(launchConfig, null, 4)
+    );
+}
+
+/**
+ * Creates all necessary project files with template content
+ */
+async function createProjectFiles(projectPath: string, projectName: string) {
```
TwitchBronBron, line 209: Let's move the Roku project files out of typescript strings and instead store them as actual files in the project. Maybe call the folder `project-templates`. Then you can do a simple `fastGlob` lookup to find all files in there recursively and copy them to the `projectPath`.

Then for inserting the `projectName` and other variables, let's just establish some standard names that we can do a quick replacement on as we copy them. (like `title=VscodeTemplateProjectName`)

This has a few benefits:
 - it's easier to scale out more starter projects 
 - the projects will actually be runnable on their own from their folders
 - it's easier to maintain compared to strings

images/channel-poster_fhd.png
TwitchBronBron, line 1: I think I'd like images that aren't "roku" branded for these template projects. We're not an official Roku product so there might be licensing/copyright concerns with distributing these images.

TwitchBronBron (changes requested):
Hey, this is looking pretty sweet! Thanks for your work on this. Here are a few suggestions.

Issue #397, opened 2022-05-19
Wrong colors in AA function names


chrisdp:
![image](https://user-images.githubusercontent.com/9591618/169376658-fc2eae65-6dd0-4dd0-8369-873f0510e35c.png)

sethmaclean:
@TwitchBronBron 
PR: https://github.com/rokucommunity/vscode-brightscript-language/pull/627 was approved & merged which was opened to address this bug.

This bug should be ready to close.

PR #627, opened 2025-03-04
bugFix/syntax highlighting on AA function names

sethmaclean:
This PR is linked to issue #397 

It was discovered that the `support-component-function` classes were applying text-mate styles to functions that should not be receiving them. reserved function names already receive syntax highlighting following the regex colourization, by the semantic styling in bright-script. as a result, all references to the support component class objects have been removed from the tmLanguage.bs file. 

Unit tests have been added to validate the highlighting of functions and variables named that match reserved function names in Roku.


TwitchBronBron (changes requested):
Thanks for this. Can you add some unit tests to verify this is working correctly (and ensure we don't break it again in the future)?

PR #628, opened 2025-03-04
bug fix syntax highlighting on return statements

sethmaclean:
This bug fix is a follow up to https://github.com/rokucommunity/vscode-brightscript-language/pull/627

I identified that return statements in function calls were receiving the incorrect syntax highlighting (yellow) due to the order of operations in which they are processed.

The order has been updated to account for the very first position that `keyword_return` received the correct syntax highlighting within the repo-patterns hierarchy.

### DIFF
--- a/syntaxes/brightscript.tmLanguage.json
+++ b/syntaxes/brightscript.tmLanguage.json
@@ -134,10 +134,10 @@
                     "include": "#operators"
                 },
                 {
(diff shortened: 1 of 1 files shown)


TwitchBronBron (changes requested):
Can you add some unit tests to verify this is working properly and ensure it doesn't break in the future?

PR #629, opened 2025-03-05
syntax highlighting keyword modifier class

sethmaclean:
This PR adds the keyword modifier class for syntax highlighting when the "optional" keyword is used in interface declarations. 

Note: There may very well be a separate class that we would like to store this value in instead. if so please LMK and I will make the necessary adjustments.

Before:
image

After:
image

### DIFF
--- a/syntaxes/brightscript.tmLanguage.json
+++ b/syntaxes/brightscript.tmLanguage.json
@@ -134,10 +134,10 @@
                     "include": "#operators"
                 },
                 {
-                    "include": "#support_functions"
+                    "include": "#variables_and_params"
                 },
                 {
-                    "include": "#variables_and_params"
+                    "include": "#support_functions"
                 },
                 {
                     "include": "#annotation"
--- a/syntaxes/brightscript.tmLanguage.json
+++ b/syntaxes/brightscript.tmLanguage.json
@@ -80,6 +80,9 @@
                     "include": "#keyword_logical_operator"
                 },
                 {
+                    "include": "#keyword_return"
+                },
+                {
                     "include": "#function_call"
                 },
                 {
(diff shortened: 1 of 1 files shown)


syntaxes/brightscript.tmLanguage.json
```diff
@@ -781,7 +797,7 @@
             }
         },
         "interface_field": {
-            "begin": "(?i)\\s*\\b([a-z0-9_]+)(?:[\\s\\t]*(as))?",
+            "begin": "(?i)\\s*\\b([a-z0-9_]+)(?:[\\s\\t]*(as))",
```
TwitchBronBron, line 800: I think we might want to add  `optional` in this regex as an additional optional capture group rather than capturing it as its unique named item.

syntaxes/brightscript.tmLanguage.json
```diff
@@ -79,6 +79,9 @@
                 {
                     "include": "#keyword_logical_operator"
                 },
+                {
+                    "include": "#keyword_modifier"
```
TwitchBronBron, line 83: Adding this here would probably result in certain variables or properties named `optional` being incorrectly colored. Like:
```
sub optional(optional = true)
    optional = false
    return optional
end sub
```

TwitchBronBron (changes requested):
I have a few suggestions. Also, can you add some unit tests to verify the colorization works in interface declarations, and also that we are _not_ capturing `optional` as blue in other situations (like function names, parameters, and variables).

syntaxes/brightscript.tmLanguage.json
sethmaclean: removed the `keyword_modifier `AA & all references to it, as 'optional' is being handled by `class_or_interface_field`

sethmaclean:
still not working quite right, the following fields are not styling properly due to overlapping issues with the function logic in entire_language
image

PR #630, opened 2025-03-10
Reenable the `deferScopeLoading` launch option

TwitchBronBron:
Reenable the `deferScopeLoading` launch option now that the underlying crash has been fixed.


TwitchBronBron:
oops. This PR was definitely named incorrectly. Should have been the `enhanceREPLCompletions` option. lol oh well. nobody reads the commit history anyway, right?

PR #644, opened 2025-05-20
Added support for Exporting and Importing automation scripts.

timbyu:
Added support for Exporting and Loading automation scripts. It exports and loads all scripts, not individual scripts.

This is my first attempt at adding code to a VS Code extention. I am very open to suggestions and corrections.

### DIFF
--- a/src/viewProviders/RokuAutomationViewViewProvider.ts
+++ b/src/viewProviders/RokuAutomationViewViewProvider.ts
@@ -51,7 +51,7 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
         this.registerCommand(VscodeCommand.rokuAutomationViewImportAllAutomations, async () => {
             if (this.isRecording) {
                 // Only allow importing when we aren't currently recording
-                vscode.window.showInformationMessage('Cannot import automation scripts while recording. Please stop recording first.');
+                void vscode.window.showInformationMessage('Cannot import automation scripts while recording. Please stop recording first.');
                 return;
             }
 
@@ -75,7 +75,7 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
                 },
                 defaultUri: vscode.Uri.file('automation.json'),
                 canSelectMany: false
-            })
+            });
 
             if (!filePath) {
                 return;
@@ -96,18 +96,18 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
 
                 this.updateCurrentRunningStep();
                 this.onImportAllAutomations();
-                vscode.window.showInformationMessage('Automation scripts imported successfully from ' + filePath[0].fsPath);
+                void vscode.window.showInformationMessage('Automation scripts imported successfully from ' + filePath[0].fsPath);
             } catch (err) {
-                vscode.window.showErrorMessage('Failed to import automations: ' + (err as Error).message);
+                void vscode.window.showErrorMessage('Failed to import automations: ' + (err as Error).message);
             }
         });
 
         this.registerCommand(VscodeCommand.rokuAutomationViewExportAllAutomations, async () => {
-            vscode.window.showInformationMessage('Exporting automation data...');
+            void vscode.window.showInformationMessage('Exporting automation data...');
 
             if (this.isRecording) {
                 // Only allow exporting when we aren't currently recording
-                vscode.window.showInformationMessage('Cannot export automation scripts while recording. Please stop recording first.');
+                void vscode.window.showInformationMessage('Cannot export automation scripts while recording. Please stop recording first.');
                 return;
             }
 
@@ -127,7 +127,7 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
                     'JSON': ['json'],
                     'All Files': ['*']
                 },
-                defaultUri: defaultUri,
+                defaultUri: defaultUri
             });
 
             if (!filePath) {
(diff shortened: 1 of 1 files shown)


src/viewProviders/RokuAutomationViewViewProvider.ts
```diff
@@ -9,6 +9,7 @@ import { ViewProviderId } from './ViewProviderId';
 import { ViewProviderCommand } from './ViewProviderCommand';
 import { ViewProviderEvent } from './ViewProviderEvent';
 import { WorkspaceStateKey } from './WorkspaceStateKey';
+import { readFileSync, writeFileSync } from 'fs';
```
triwav, line 12: @TwitchBronBron any thought on importing this way?

src/viewProviders/RokuAutomationViewViewProvider.ts
```diff
@@ -47,6 +48,85 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
...
+        });
+
+        this.registerCommand(VscodeCommand.rokuAutomationViewExportAutomation, async () => {
+            vscode.window.showInformationMessage('Exporting automation data...');
+
+            if (this.isRecording) {
+                // Only allow exporting when we aren't currently recording
+                await this.setIsRecording(false);
+            }
+            vscode.window.showSaveDialog({
```
triwav, line 110: Optional: You could write this as `const filePath = await vscode.window.showSaveDialog({` to make the code more readable

src/viewProviders/RokuAutomationViewViewProvider.ts
TwitchBronBron: I don't have any fundamental issues with this pattern, but I think we import as `fs` everywhere else in the codebase so I'd say let's do that for consistency sake. 

Also, let's use `fs-extra` instead which comes with a few extra protections. 

```suggestion
import * as fsExtra from 'fs-extra';
```

Then usage, you can use `.outputFileSync` instead of `writeFileSync` so it auto-createes directories. 

https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/outputFile.md

package.json
```diff
@@ -374,6 +374,17 @@
                     "when": "view == rokuAutomationView && brightscript.rokuAutomationView.autorunOnDeploy",
                     "group": "navigation@2"
                 },
+                {
+                    "command": "extension.brightscript.rokuAutomationView.exportAutomation",
+                    "when": "view == rokuAutomationView",
+                    "group": "navigation@2"
+                },
+                {
+                    "command": "extension.brightscript.rokuAutomationView.loadAutomation",
```
TwitchBronBron, line 383: should this command name be `import` instead of "load" to follow the typical "import/export" language?

src/viewProviders/RokuAutomationViewViewProvider.ts
```diff
@@ -47,6 +48,85 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
             }
         });
 
+        this.registerCommand(VscodeCommand.rokuAutomationViewLoadAutomation, async () => {
+            if (this.isRecording) {
+                // Only allow loading when we aren't currently recording
+                await this.setIsRecording(false);
```
TwitchBronBron, line 54: Hmm. I wonder if we should show a "can't load while recording" error rather than cancelling recording?

src/viewProviders/RokuAutomationViewViewProvider.ts
```diff
@@ -47,6 +48,85 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
...
+                    'This will replace all currently loaded automation scripts. Continue?',
+                    { modal: true },
+                    'Yes'
+                );
+                if (confirm !== 'Yes') {
+                    return;
+                }
+            }
+
+            vscode.window.showOpenDialog({
```
TwitchBronBron, line 69: Can you fix these warnings? I think you need to await the dialogs.

src/viewProviders/RokuAutomationViewViewProvider.ts
```diff
@@ -47,6 +48,85 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
...
+                }
+            });
+        });
+
+        this.registerCommand(VscodeCommand.rokuAutomationViewExportAutomation, async () => {
+            vscode.window.showInformationMessage('Exporting automation data...');
+
+            if (this.isRecording) {
+                // Only allow exporting when we aren't currently recording
+                await this.setIsRecording(false);
```
TwitchBronBron, line 108: Same as the other function, we should cancel the action with a warning popup instead of disabling recording.

src/viewProviders/RokuAutomationViewViewProvider.ts
```diff
@@ -47,6 +48,85 @@ export class RokuAutomationViewViewProvider extends BaseRdbViewProvider {
...
+            }
+            vscode.window.showSaveDialog({
+                title: 'Export Automation scripts',
+                filters: {
+                    'JSON': ['json'],
+                    'All Files': ['*']
+                },
+                defaultUri: vscode.Uri.file('automation.json')
+            }).then(async (filePath) => {
```
TwitchBronBron, line 117: Can you convert this to async code using `await` instead of using promises.

TwitchBronBron (changes requested):
This is looking pretty good so far. Got a few suggestions.

src/viewProviders/RokuAutomationViewViewProvider.ts
timbyu: updated

TwitchBronBron (approved):

PR #657, opened 2025-07-04
Add new launch variable substitution called `activeHost`

jeanbenitezu:
This PR introduces a new variable substitution, `${activeHost}`, that allows users to dynamically reference the currently configured remote host in their debug configurations.

In addition, I added a small documentation about this variable substitution functionality, which may help others identify their goal.

🚀 Why
- Enables more flexible debug configurations that can adapt to different remote hosts
- Reduces the need to manually update host references across multiple configuration files

### DIFF
--- a/src/DebugConfigurationProvider.spec.ts
+++ b/src/DebugConfigurationProvider.spec.ts
@@ -53,7 +53,8 @@ describe('BrightScriptConfigurationProvider', () => {
             activeDeviceManager,
             null,
             vscode.window.createOutputChannel('Extension'),
-            userInputManager
+            userInputManager,
+            null // BrightScriptCommands is not used in this test (help!)
         );
     });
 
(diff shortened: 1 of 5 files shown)


jeanbenitezu:
Hey Bron!

Thanks so much for taking the time to review this. I really appreciate it.

I was actually on the fence about the variable name, so `activeHost` sounds perfect, much cleaner and more intuitive 😅 I’ll make that change and push it shortly.

Also, the upcoming device picker feature sounds awesome!! looking forward to seeing it in action!

Happy Friday!

src/DebugConfigurationProvider.spec.ts
```diff
@@ -53,7 +53,8 @@ describe('BrightScriptConfigurationProvider', () => {
             activeDeviceManager,
             null,
             vscode.window.createOutputChannel('Extension'),
-            userInputManager
+            userInputManager,
+            null // BrightScriptCommands is not used in this test (help!)
```
TwitchBronBron:
```suggestion
            null // BrightScriptCommands is not used in this test
```

`null` is fine here. No need to do more work than the test requires.

PR #659, opened 2025-07-18
Support actions on the popup message event

Christian-Holbrook:


src/LogOutputManager.ts
```diff
@@ -229,7 +229,9 @@ export class LogOutputManager {
             this.appendLine(e.body.line);
 
         } else if (isPopupMessageEvent(e)) {
-            this.showMessage(e.body.message, e.body.severity, e.body.modal);
+            this.showMessage(e.body.message, e.body.severity, e.body.modal, ...e.body.actions).catch((err) => {
```
TwitchBronBron: Should probably guard against e.body.actions not existing. 
```suggestion
            this.showMessage(e.body.message, e.body.severity, e.body.modal, ...(e?.body?.actions ?? [])).catch((err) => {
```

package.json
TwitchBronBron, line 1: need to revert these. Probably also the package-lock.json

PR #677, opened 2025-12-15
Customize screenshot directory path

### LINKED ISSUE #539: Customize screenshot directory path
It would be nice to allow developers to override the default screenshot directory, as right now it uses the OS's temp dir.

- Define a new user setting called `brightscript.screenshotDir` [here](https://github.com/rokucommunity/vscode-brightscript-language/blob/701b00a224f719b1fd0907eb32d46470b8de6623/package.json#L2386) in the miscellaneous group.
- You can retrieve the setting with this code: 
     ```
     vscode.workspace.getConfiguration('brightscript').get<string>('screenshotDir');
    ````
-  We capture the screenshot [here](https://github.com/rokucommunity/vscode-brightscript-language/blob/701b00a224f719b1fd0907eb32d46470b8de6623/src/commands/CaptureScreenshotCommand.ts#L38). I think you need to set `outDir` as a property in that object that is passed to `rokuDeploy.takeScreenshot()`

hasimk-dna:
fixes #539

### DIFF
--- a/src/BrightScriptTaskProvider.spec.ts
+++ b/src/BrightScriptTaskProvider.spec.ts
@@ -674,10 +674,9 @@ describe('BrightScriptTaskProvider', () => {
             sinon.restore();
         });
 
-        it('returns Windows shell configuration on win32', async () => {
+        it('uses cmd.exe on Windows with env from terminal.integrated.env.windows', async () => {
             Object.defineProperty(process, 'platform', { value: 'win32' });
 
-            configStub.get.withArgs('shell.windows').returns('powershell.exe');
             configStub.get.withArgs('env.windows').returns({ CUSTOM_VAR: 'value' });
 
             const task = createMockTask({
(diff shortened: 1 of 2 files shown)


src/commands/CaptureScreenshotCommand.ts
```diff
@@ -1,53 +1,81 @@
...
-                        void vscode.commands.executeCommand('vscode.open', vscode.Uri.file(screenshotPath));
+    private async captureScreenshot(hostParam?: string) {
+        const { host, password } = await this.getHostAndPassword(hostParam);
+
+        await vscode.window.withProgress({
+            title: `Capturing screenshot from '${host}'`,
+            location: vscode.ProgressLocation.Notification
+        }, async () => {
+            try {
+                let screenshotDir = vscode.workspace.getConfiguration('brightscript').get<string>('screenshotDir');
```
TwitchBronBron: Could we move this into a separate private method, maybe called `getScreenshotDir()`

PR #679, opened 2026-01-06
Perfetto controls

JyotiBurnwal1:
Support for enable, start and stop perfetto feature.


package.json
```diff
@@ -910,6 +922,32 @@
                                 "description": "If true, will terminate the debug session if app exit is detected. This option is ignored when enableDebugProtocol is set to true. This currently relies on 9.1+ launch beacon notifications, so will not work on a pre 9.1 device.",
                                 "default": true
                             },
+                            "profiling": {
+                                "type": "object",
+                                "description": "Configuration for profiling functionality",
+                                "default": {
+                                    "enable": false,
+                                    "dir": "${workspaceFolder}/profiles",
+                                    "filename": "${appTitle}_${timestamp}.prof"
```
ramkav: please use ".perfetto-trace" instead of ".prof" for the filename suffix.

package.json
```diff
@@ -910,6 +922,32 @@
...
+                                    },
+                                    "dir": {
+                                        "type": "string",
+                                        "description": "Directory where profile files should be stored",
+                                        "default": "${workspaceFolder}/profiles"
+                                    },
+                                    "filename": {
+                                        "type": "string",
+                                        "description": "The name of the profile file. Can include variables like ${appTitle} and ${timestamp}",
+                                        "default": "${appTitle}_${timestamp}.prof"
```
ramkav: Please use ".perfetto-trace" instead of ".prof" for filename suffix.

package.json
```diff
@@ -2413,6 +2451,33 @@
...
+                            },
+                            "dir": {
+                                "type": "string",
+                                "description": "Directory where profile files should be stored",
+                                "default": "${workspaceFolder}/profiles"
+                            },
+                            "filename": {
+                                "type": "string",
+                                "description": "The name of the profile file. Can include variables like ${appTitle} and ${timestamp}",
+                                "default": "${appTitle}_${timestamp}.prof"
```
ramkav: Please use ".perfetto-trace" instead of ".prof" for filename suffix.

src/DebugConfigurationProvider.ts
```diff
@@ -63,6 +63,11 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
         showHiddenVariables: false,
         enableDebuggerAutoRecovery: false,
         stopDebuggerOnAppExit: false,
+        profiling: {
+            enable: false,
+            dir: '${workspaceFolder}/profiles/',
+            filename: '${appTitle}-${timestamp}.bsprof'
```
ramkav: Please ".perfetto-trace" instead of ".bsprof" for trace filename suffix.
Also, later in the code when testing for profiling, what should the default behavior be if the filename and/or dir are missing? What's the default?

src/SceneGraphDebugCommands.ts
```diff
@@ -166,6 +183,64 @@ export class SceneGraphDebugCommands {
...
+        const perfettoController = new PerfettoControls(this.host);
+        
+        switch (command) {
+            case 'start':
+                this.showInfo('Starting perfetto Tracing!');
+                
+                const startResponse = await perfettoController.startTracing();
+                
+                if (startResponse.error) {
+                    this.showError('Error starting perfetto tracing: ' + startResponse.message);
```
ramkav: Can the error messages be made more consistent? In some cases we say "perfetto tracing", in others, it's "Perfetto tracing, and in other places it's "Tracing" or "Perfetto tracing". If someone's doing a case-sensitive filter, this can lead to a needlessly painful debugging experience. 

Also, perhaps a good sign that this logic needs to be made a bit more DRY? Maybe use a Template Method pattern? Im thinking something like this (so we have only a couple places where the messages are printed, reducing the chance for variability, and making the code a bit tighter?):

private async handleTracingOperation(
        operationName: string,
        operation: () => Promise<PerfettoResponse>
    ): Promise<void> {
        this.showInfo(`${operationName} perfetto Tracing!`);

        const response = await operation();

        if (response.error) {
            this.showError(`Error ${operationName.toLowerCase()} perfetto tracing: ${response.message}`);
        } else {
            this.showInfo(`Perfetto tracing ${operationName.toLowerCase()}d: ${response.message}`);
        }
    }

    public async executeCommand(command: string): Promise<void> {
        switch (command) {
            case 'start':
                await this.handleTracingOperation('Starting', () => this.perfettoController.startTracing());
                break;
            case 'stop':
                await this.handleTracingOperation('Stopping', () => this.perfettoController.stopTracing());
                break;
            case 'enable':
                await this.handleTracingOperation('Enabling', () => this.perfettoController.enableTracing());
                break;
            // Add other cases here
            default:
                this.showError(`Unknown command: ${command}`);
                break;
        }
    }
}

src/commands/RekeyAndPackageCommand.ts
```diff
@@ -369,6 +369,13 @@ export class RekeyAndPackageCommand {
         }
         rokuDeployOptions.packageConfig = 'launch.json: ' + selectedConfig.rootDir;
 
+        if (selectedConfig?.profiling?.dir.includes('${workspaceFolder}')) {
```
ramkav: What should happen if profiling is enabled but there's no dir? Does the DebugConfigurationProvider.ts value become the default for any missing fields? E.g. if the developer's launch.json has:
profiling: { enable: true}

will the dir and filename be filled by default from DebugConfigurationProvider.ts as:
profiling: {
            enable: true, <--- this is from developer's launch.json
            dir: '${workspaceFolder}/profiles/', <--- populated from DebugConfigurationProvider.ts?
            filename: '${appTitle}-${timestamp}.bsprof'  <--- populated from DebugConfigurationProvider.ts?
}

package.json
JyotiBurnwal1: done

src/DebugConfigurationProvider.ts
JyotiBurnwal1: done

PR #680, opened 2026-01-14
Perfetto tracing

JyotiBurnwal1:
### Supported Config

```json
{
  "profiling": {
    "tracing": {
      "enable": true,
      "dir": "${workspaceFolder}/profiling",
      "filename": "${appTitle}_${timestamp}_${sequence}.perfetto-trace",
      "connectOnStart": false
    }
  }
}

Notes:

- If both enable and connectOnStart are true, connectOnStart takes precedence.
- If both ${timestamp} and ${sequence} are provided in the filename, only ${timestamp} is used to ensure unique trace files.
- Wait for the app to fully launch on FW before clicking Start Tracing.
- The heap snapshot button will appear only after an active tracing.

### DIFF
--- a/src/util.spec.ts
+++ b/src/util.spec.ts
@@ -1,8 +1,6 @@
 import * as assert from 'assert';
 import { expect } from 'chai';
 import * as fsExtra from 'fs-extra';
-import * as getPort from 'get-port';
-import * as net from 'net';
 import * as path from 'path';
 import * as sinonActual from 'sinon';
 

### CODE REVIEW

File: src/commands/RekeyAndPackageCommand.ts [hunk 1]
```diff
@@ -326,6 +326,12 @@ export class RekeyAndPackageCommand {
         }
     }
 
+    private async workspacePath() {
+        await this.brightScriptCommands.getWorkspacePath();
+        let workspacePath = this.brightScriptCommands.workspacePath;
+        return workspacePath;
```
TwitchBronBron: `.getWorkspacePath()` actually returns the path, so no need to read it off the property. However, I don't see a need to duplicate that function, just read it from brightscriptCommands down below. So let's just delete this code.


```suggestion
        let workspacePath = await this.brightScriptCommands.getWorkspacePath();
        return workspacePath;
```

src/commands/RekeyAndPackageCommand.ts
```diff
@@ -362,13 +368,22 @@ export class RekeyAndPackageCommand {
         }
 
         if (selectedConfig.rootDir?.includes('${workspaceFolder}')) {
-            await this.brightScriptCommands.getWorkspacePath();
-            let workspacePath = this.brightScriptCommands.workspacePath;
-
+            let workspacePath = await this.workspacePath();
```
TwitchBronBron: `.getWorkspacePath()` returns the value, so we don't need to mirror that with our own getter. 

```suggestion
            let workspacePath = await this.brightScriptCommands.getWorkspacePath()
```

package.json
```diff
@@ -910,6 +946,32 @@
...
+                            "profiling": {
+                                "type": "object",
+                                "description": "Configuration for profiling functionality",
+                                "default": {
+                                    "enable": false,
+                                    "dir": "${workspaceFolder}/profiles",
+                                    "filename": "${appTitle}_${timestamp}.perfetto-trace"
+                                },
+                                "properties": {
+                                    "enable": {
```
TwitchBronBron: Let's call this `enabled` (with a trailing d). That aligns with the same pattern from `fileLogging` 
```suggestion
                                    "enabled": {
```

package.json
```diff
@@ -910,6 +946,32 @@
...
+                                        "type": "boolean",
+                                        "description": "Whether profiling is enabled",
```
TwitchBronBron:
```suggestion
                                        "description": "Should profiling be enabled",
```

src/PerfettoControlCommands.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    private async perfettoControls(command: string) {
+        await this.getRemoteHost();
+        const perfettoController = new PerfettoControls(this.host);
+        
+        switch (command) {
+            case 'start':
+                const startFilename = "something.perfetto-trace";
+                const startFolderUri = await this.createFolderIfNotExists("perfetto");
```
TwitchBronBron: Let's use the fsExtra class for this, no need for a custom function. (apply to all the places where you call createFolderIfNotExists)
```suggestion
                const startFolderUri = fsExtra.ensureDirSync("perfetto");
```

src/PerfettoControlCommands.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    private showInfo(msg: string) {
+        vscode.window.showInformationMessage(msg);
+    }
+
+    private showError(msg: string) {
+        vscode.window.showErrorMessage(msg);
+    }
+
+    private async createFolderIfNotExists(folderName: string): Promise<vscode.Uri> {
```
TwitchBronBron: I don't think you need this function. Just run `fsExtra.ensureDirSync`.

src/PerfettoControlCommands.ts
```diff
@@ -0,0 +1,243 @@
...
+            throw new Error(errorMessage);
+        }
+
+        const folderDir = vscode.Uri.joinPath(workspaceFolders[0].uri, folderName);
+        await vscode.workspace.fs.createDirectory(folderDir);
+        
+        return folderDir;
+    }
+
+    public async getRemoteHost() {
```
TwitchBronBron: Don't just copy-paste this function from somewhere else. You can call the existing function from an instance of `BrightScriptCommands`. (`this.brightScriptCommands.getRemoteHost()`).

src/extension.ts
```diff
@@ -180,6 +181,8 @@ export class Extension {
 
         //register all commands for this extension
         this.brightScriptCommands.registerCommands();
+        vscode.window.showErrorMessage(JSON.stringify(Object.keys(context)));
+        perfettoControlCommands.registerPerfettoControlCommands(context);
```
TwitchBronBron: Let's move this inside of `BrightScriptCommands.registerCommands`

src/extension.ts
```diff
@@ -180,6 +181,8 @@ export class Extension {
 
         //register all commands for this extension
         this.brightScriptCommands.registerCommands();
+        vscode.window.showErrorMessage(JSON.stringify(Object.keys(context)));
```
TwitchBronBron: Remove this alert.

src/PerfettoControlCommands.ts
```diff
@@ -0,0 +1,243 @@
...
+        subscriptions.push(vscode.commands.registerCommand('extension.brightscript.stopTracing', async () => {
+            await this.perfettoControls("stop");
+            await vscode.commands.executeCommand(
+                'setContext',
+                'brightscript.tracingActive',
+                false
+            );
+        }));
+
+        subscriptions.push(
```
TwitchBronBron: I don't think this editor provider registration belongs in this class. Instead, it should be placed in `extension.ts` instead. Maybe close to where we call ` vscode.window.registerTreeDataProvider`

src/PerfettoControlCommands.ts
```diff
@@ -0,0 +1,243 @@
...
+            {
+                webviewOptions: {
+                retainContextWhenHidden: true, // Optional: Keep webview state when hidden
+                },
+            }
+            )
+        );
+    }
+
+    private async perfettoControls(command: string) {
```
TwitchBronBron: I'm ont sure there's much benefit to having a switch case approach here. It may be cleaner to break each of these into standalone private methods, that way we get better intellisense. 

Something like this (names up for discussion)

```
this.startPerfetto();
this.stopPerfetto();
this.enablePerfetto();
```

Also add jsdoc comments explaining what each of them do.

src/DebugConfigurationProvider.ts
```diff
@@ -63,6 +63,11 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
         showHiddenVariables: false,
         enableDebuggerAutoRecovery: false,
         stopDebuggerOnAppExit: false,
+        profiling: {
+            enable: false,
+            dir: '${workspaceFolder}/profiles/',
+            filename: '${appTitle}-${timestamp}.perfetto-trace'
```
TwitchBronBron: Let's also support `${sequence}`, which would look at all the other filename info, read disk to find the last sequence, and then increment. For example:

file-1.log
file-2.log
next would 3

src/commands/RekeyAndPackageCommand.ts
JyotiBurnwal1: done

src/commands/RekeyAndPackageCommand.ts
JyotiBurnwal1: done

package.json
JyotiBurnwal1: done

package.json
JyotiBurnwal1: done

src/PerfettoControlCommands.ts
JyotiBurnwal1: done but this is not working as expected. This is WIP.

src/PerfettoControlCommands.ts
JyotiBurnwal1: removed

src/PerfettoControlCommands.ts
JyotiBurnwal1: removed and used the suggested snippet

src/extension.ts
JyotiBurnwal1: done

src/extension.ts
JyotiBurnwal1: done

src/PerfettoControlCommands.ts
JyotiBurnwal1: removed

src/PerfettoControlCommands.ts
JyotiBurnwal1: done

src/DebugConfigurationProvider.ts
JyotiBurnwal1: this is WIP and would be implemented in next PR

JyotiBurnwal1: done

JyotiBurnwal1:
added heap snapshot functionality merged JyotiBurnwal1:heap-snapshot

src/ProfilingTracingIntegration.spec.ts
```diff
@@ -0,0 +1,487 @@
...
+
+        // Stub the vscodeContextManager.set method
+        contextManagerSetStub = sinon.stub(vscodeContextManager, 'set').resolves();
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    describe('TC-01: Tracing Disabled - Buttons Hidden', () => {
```
TwitchBronBron: Let's remove these TC style prefixes. Those don't really mean anything to this project.

JyotiBurnwal1: done

PR #683, opened 2026-02-03
Device Discovery Improvements

Christian-Holbrook:
This pull request makes significant changes to device discovery and management, focusing on refactoring the device management system, updating dependencies, and improving the user interface and command structure for device-related actions in the extension. The most notable updates are the removal of the old `ActiveDeviceManager` in favor of a new `DeviceManager`, cleaning up dependencies, and enhancing the VS Code UI/UX for device operations.

**Device Management Refactor and UI Improvements:**

* Replaced the legacy `ActiveDeviceManager` with a new `DeviceManager` implementation, updating all references and relevant logic to use the new system. This includes updating command registrations and their handlers in `BrightScriptCommands` and related test files. [[1]](diffhunk://#diff-85aa55a2b674aae8a4a886dec600426d8dad32363e7f58a88568a75343a106d5L1-L372) [[2]](diffhunk://#diff-4bc1080c0975fa8a422efbc8f73e8e5631b4a726a5fdb30402536542a820c553L13-R13) [[3]](diffhunk://#diff-4bc1080c0975fa8a422efbc8f73e8e5631b4a726a5fdb30402536542a820c553L26-R26) [[4]](diffhunk://#diff-f4049619d65dac28265033c5b5b6bd3d96caab2cacf0dff3ed14b6b96f9bde69L12-R13)
* Added new device management commands: `refreshDevice`, `setActiveDevice`, `clearDeviceCache`, and `clearLastSeenDevices`, with corresponding UI elements and context menu actions in the VS Code extension. These allow users to refresh device lists, set the active device directly from the UI, and clear cached device data. [[1]](diffhunk://#diff-4bc1080c0975fa8a422efbc8f73e8e5631b4a726a5fdb30402536542a820c553L55-R67) [[2]](diffhunk://#diff-4bc1080c0975fa8a422efbc8f73e8e5631b4a726a5fdb30402536542a820c553R331-R341) [[3]](diffhunk://#diff-4bc1080c0975fa8a422efbc8f73e8e5631b4a726a5fdb30402536542a820c553L385-R422) [[4]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R2838-R2849) [[5]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R3101-R3112) [[6]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R379-R390)

**Dependency Cleanup:**

* Removed unused or obsolete dependencies from `package.json`, including `backoff`, `fast-xml-parser`, `net`, and their associated type definitions, as well as `@types/request` and `@types/backoff`. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L57-L67) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L76) [[3]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L98)

**VS Code View and Command Updates:**

* Renamed the device view from `onlineDevicesView` to `devicesView` and updated all related command `when` clauses and view identifiers for consistency and clarity in the UI. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L169-R164) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L250-R245)

These changes collectively modernize the device management subsystem, streamline dependencies, and provide a more user-friendly and maintainable extension interface.


src/GlobalStateManager.ts
```diff
@@ -68,6 +68,34 @@ export class GlobalStateManager {
         }
     }
 
+    public getKnownDeviceIpsByNetwork(network: string): string[] {
+        const networks = this.context.globalState.get<Record<string, string[]>>(this.keys.knownDeviceIpsByNetwork) || {};
```
TwitchBronBron: Let's clean up any networks that have expired (based on their "lastUpdatedDate" prop), but do it at the end, so if we were going to clean up the current network, we still pass those in.

src/GlobalStateManager.ts
```diff
@@ -68,6 +68,34 @@ export class GlobalStateManager {
...
+        return networks[network] || [];
+    }
+
+    public setKnownDeviceIpsByNetwork(network: string, ips: string[]) {
```
TwitchBronBron: Should only run this function (externally) when the editor is focused

src/deviceDiscovery/NetworkChangeMonitor.ts
```diff
@@ -0,0 +1,114 @@
...
+                if (
+                    net.family === 'IPv4' &&
+                    !net.internal &&
+                    !net.address.startsWith('169.254.') // link-local
+                ) {
+                    ips.add(net.address);
+                }
+            }
+        }
+
```
TwitchBronBron: prolly can scrap this?

src/deviceDiscovery/NetworkChangeMonitor.ts
```diff
@@ -0,0 +1,114 @@
...
+        // Check if IPs changed
+        const added = [...currentIps].filter(ip => !this.previousIps.has(ip));
+        const removed = [...this.previousIps].filter(ip => !currentIps.has(ip));
+
+        if (added.length > 0 || removed.length > 0) {
+            this.onNetworkChanged();
+        }
+
+        this.previousIps = currentIps;
```
TwitchBronBron: could just do something like `if(this.currentHash <> getNetworkHash()){ emitEvent(); }`

src/deviceDiscovery/NetworkChangeMonitor.ts
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    public onFocusGain(): void {
+        this.start();
+    }
+
+    public onFocusLost(): void {
+        this.stop();
+    }
```
TwitchBronBron: Maybe we can scrap these, in favor of just having the outer `onFocusChange` event call the `.start()` and `.stop()`

src/deviceDiscovery/SystemSleepMonitor.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Monitor for system sleep/wake events by detecting gaps in timer execution
+ */
+export class SystemSleepMonitor {
+
+    private timer: NodeJS.Timeout | null = null;
+    private lastExecutionTime = 0;
+    private interval = 60000; // 1 minute
+    private gapThreshold = 120000; // 2 minutes
```
TwitchBronBron: use the underscores to separate the thousand's places. `60_000`, `120_000` etc

src/deviceDiscovery/SystemSleepMonitor.ts
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+
+    private scheduleNext(): void {
+        this.timer = setTimeout(() => {
+            if (Date.now() - this.lastExecutionTime > this.gapThreshold) {
+                this.onSleepDetected();
+            }
+            this.lastExecutionTime = Date.now();
+            this.scheduleNext();
```
TwitchBronBron: doing something like this avoids the lastExecutionTime getting stuck old forever.
```suggestion
        this.timer = setTimeout(() => {
            const lastExecutionTime = this.lastExecutionTime;
            this.lastExecutionTime = Date.now();
            this.scheduleNext();
            if (Date.now() - lastExecutionTime > this.gapThreshold) {
                this.onSleepDetected();
            }
```

src/managers/UserInputManager.ts
```diff
@@ -120,13 +155,29 @@ export class UserInputManager {
                         deferred.resolve({ manual: true });
                     } else {
                         const device = (selectedItem as any).device as RokuDeviceDetails;
+                        const isHealthy = await this.activeDeviceManager.checkDeviceHealth(device);
```
TwitchBronBron: if device is NOT healthy, ensure the picker remains open. Also add a comment explaining that is the expected behavior

src/managers/UserInputManager.ts
```diff
@@ -120,13 +155,29 @@ export class UserInputManager {
...
                     }
                     quickPick.dispose();
                 }
             }
         });
+
+        quickPick.onDidTriggerButton(button => {
+            if (button.tooltip === 'Scan for devices') {
+                quickPick.busy = true;
+                this.activeDeviceManager.discoverAll();
```
TwitchBronBron: Let's switch to calling `.refresh()` instead, which will health-check all deviecs, and also scan for new ones.

src/deviceDiscovery/ActiveDeviceManager.ts
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    /**
+     * Discover all Roku devices on the network and watch for new ones that connect
+     */
+    public discoverAll() {
+        this.needsFutureBroadcast = false;
+        this.lastBroadcastDate = new Date();
+        this.finder.scan();
+    }
```
TwitchBronBron: let's keep the "should I do a broadcast" logic inside this function, sothe outside just can pass `force` if they require it.
```suggestion
    public discoverAll(force = false) {
        this.needsFutureBroadcast = false;
        if(this.needsFutureBroadcast || this.itsBeenLongEnoughToDoABroadcast() || force === true){
            this.lastBroadcastDate = new Date();
            this.finder.scan();
            return true;
        } else {
            return false;
        }
    }
```

src/managers/UserInputManager.ts
```diff
@@ -75,13 +87,28 @@ export class UserInputManager {
...
         }
         quickPick.show();
+
+        //set a timeout to automatically start scanning for devices after a short delay
+        scanTimeoutId = setTimeout(() => {
+            if (hasBroadcastThisSession) {
+                return;
+            }
+            this.activeDeviceManager.discoverAll();
+        }, scanTimeoutMs);
```
TwitchBronBron: Check the `didExecuteBroadcast` and then only call if false, (pass in force=true)

TwitchBronBron: Can we move this timeout up closer to the top, maybe right next to the initial broadcast call?

src/managers/UserInputManager.ts
```diff
@@ -104,13 +131,21 @@ export class UserInputManager {
         //anytime the device picker adds/removes a device, update the list
         this.activeDeviceManager.on('device-found', refreshList, disposables);
         this.activeDeviceManager.on('device-expired', refreshList, disposables);
+        this.activeDeviceManager.on('need-future-broadcast', () => {
+            hasBroadcastThisSession = true;
+            if (scanTimeoutId) {
+                clearTimeout(scanTimeoutId);
+                scanTimeoutId = null;
+            }
+            this.activeDeviceManager.discoverAll();
```
TwitchBronBron: force a broadcast here. Nice catch clearing the timeout too.

src/managers/UserInputManager.ts
```diff
@@ -30,11 +30,23 @@ export class UserInputManager {
      * Prompt the user to pick a host from a list of devices
      */
     public async promptForHost(options?: { defaultValue?: string }) {
+        let hasBroadcastThisSession = false;
```
TwitchBronBron:
```suggestion
        //returns true if a broadcast actually ran, or false if we didn't need one
        let didExecuteBroadcast = this.activeDeviceManager.broadcast();
        const timeoutId = setTimeout(()=>{
            if(didExecuteBroadcast === false){
                deviceManager.needsFutureBroadcast = true;
            }
        });
        this.activeDeviceManager.on('needFutureBroadcast-changed', () => {
            //clear the timeout id
            if (scanTimeoutId) {
                clearTimeout(scanTimeoutId);
            }
            this.activeDeviceManager.discoverAll(true);
        }, disposables);
```

src/deviceDiscovery/ActiveDeviceManager.ts
```diff
@@ -0,0 +1,479 @@
...
+     */
+    public static readonly BROADCAST_STALE_THRESHOLD_MS = 30 * 60 * 1000; // 30 minutes
+
+    private _needFutureBroadcast = false;
+
+    public firstRequestForDevices: boolean;
+    public lastUsedDevice: RokuDeviceDetails;
+    public enabled: boolean;
+
+    public get needsFutureBroadcast(): boolean {
```
TwitchBronBron: let's call it `broadcastNeeded` for now. the event could be `broadcastNeeded-changed` or something... naming is hard.

src/deviceDiscovery/ActiveDeviceManager.ts
```diff
@@ -0,0 +1,479 @@
...
+     */
+    public refresh() {
+        this.healthCheckAllDevices().catch(() => { });
+        this.discoverAll();
+    }
+
+    /**
+     * Discover all Roku devices on the network and watch for new ones that connect
+     */
+    public discoverAll() {
```
TwitchBronBron: Maybe we eliminate this function, in favor of refresh. Every time, we just run "health check every device" and also "do a broadcast"

TwitchBronBron: or make it private?

src/deviceDiscovery/ActiveDeviceManager.ts
Christian-Holbrook: I was thinking about renaming discoverAll, broadcastNeeded to be named with "scan".

discoverAll -> scanForDevices
broadcastNeeded -> scanNeeded. What do you think?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,659 @@
...
+                }
+            }
+
+            //if the `concealDeviceInfo` setting was changed, refresh the UI (no reload needed)
+            if (event.affectsConfiguration('brightscript.deviceDiscovery.concealDeviceInfo')) {
+                if (this.enabled) {
+                    this.emitDevicesChanged();
+                }
+            }
+        });
```
TwitchBronBron:
```suggestion 
      const run = (event) => {
            let config: any = vscode.workspace.getConfiguration('brightscript') || {};
            this.enabled = config.deviceDiscovery?.enabled;
            this.showInfoMessages = config.deviceDiscovery?.showInfoMessages;

            //if the `deviceDiscovery.enabled` setting was changed, start or stop monitoring
            if (event.affectsConfiguration('brightscript.deviceDiscovery.enabled')) {
                if (this.enabled) {
                    this.systemSleepMonitor.start();
                    void this.activateMonitoring();
                    this.loadLastSeenDevices();
                } else {
                    this.systemSleepMonitor.stop();
                    this.deactivateMonitoring();
                }
            }

            //if the `concealDeviceInfo` setting was changed, refresh the UI (no reload needed)
            if (event.affectsConfiguration('brightscript.deviceDiscovery.concealDeviceInfo')) {
                if (this.enabled) {
                    this.emitDevicesChanged();
                }
            }
        });
           vscode.workspace.onDidChangeConfiguration(run);
           run();
```

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,659 @@
...
+
+        this.setupConfiguration();
+        this.setupWindowFocusHandling();
+        this.setupMonitors();
+        this.initializeIfEnabled();
+    }
+
+    private setupConfiguration() {
+        let config: any = vscode.workspace.getConfiguration('brightscript') || {};
+        this.enabled = config.deviceDiscovery?.enabled;
```
TwitchBronBron: Rename this to something like `passiveScanPermitted`. It should be used for blocking any scans that were not triggered directly by user interaction (i.e. clicking the refresh buttons...)

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,659 @@
...
+            this.setScanNeeded();
+        });
+    }
+
+    private initializeIfEnabled() {
+        if (this.enabled) {
+            // Sleep monitor runs all the time when enabled (ignores focus state)
+            this.systemSleepMonitor.start();
+
+            this.loadLastSeenDevices();
```
TwitchBronBron: We should load these in the constructor at first construction, every time, (not behind a flag).

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,659 @@
...
+
+            if (vscode.window.state.focused) {
+                this.emitDevicesChanged();
+            }
+        }
+    }
+
+    private async isDeviceResponding(device: RokuDeviceDetails): Promise<boolean> {
+        try {
+            await rokuDeploy.getDeviceInfo({
```
TwitchBronBron: We're throwing this away, but anytime we do a health check (which runs a device-info query), we should use that response to update the info in the data store

TwitchBronBron: Maybe the .healthCheckDevice could just do it all. fetch deviceInfo, set new info, and flip the status as well.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,659 @@
...
+}
+
+export type DeviceState = 'offline' | 'pending' | 'online';
+
+export interface RokuDeviceDetails {
+    location: string;
+    id: string;
+    ip: string;
+    deviceState: DeviceState;
+    deviceInfo: {
```
TwitchBronBron: roku-deploy has a version of this deviceInfo data somewhere. Could you see if possibly we could eliminate this version and use that one instead?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,577 @@
...
+            ip: device.ip,
+            deviceInfo: device.deviceInfo
+        });
+
+        // Reset scan settle timer when device response comes in
+        if (this.isScanning) {
+            this.resetSettleTimer();
+        }
+
+        if (vscode.window.state.focused) {
```
TwitchBronBron: Probably don't need this check.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,577 @@
...
+    constructor(globalStateManager: GlobalStateManager) {
+        this.globalStateManager = globalStateManager;
+        this.firstRequestForDevices = true;
+        this.networkId = getNetworkHash();
+
+        this.setupConfiguration();
+        this.setupWindowFocusHandling();
+        this.setupMonitors();
+        this.initialize();
+    }
```
TwitchBronBron: can we move this to the top? I like seeing the constructor before all the props

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,567 @@
...
+            this.activateMonitoring().then(() => {
+                const lastSeenDeviceIds = this.globalStateManager.getLastSeenDeviceIds(this.networkId);
+                if (lastSeenDeviceIds.length === 0) {
+                    this.refresh();
+                } else {
+                    this.setScanNeeded();
+                }
+            }).catch(() => { });
+        }
+    }
```
TwitchBronBron, line 164:
```suggestion
    private async initialize() {
        this.loadLastSeenDevices();

        if (this.passiveScanPermitted) {
            // Sleep monitor runs all the time when enabled (ignores focus state)
            this.systemSleepMonitor.start();

            try {
                await this.activateMonitoring();
            } catch {}
            const lastSeenDeviceIds = this.globalStateManager.getLastSeenDeviceIds(this.networkId);
            if (lastSeenDeviceIds.length === 0) {
                this.refresh();
            } else {
                this.setScanNeeded();
            }
        }
    }
```

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,567 @@
...
+        return false;
+    }
+
+    private async resolveDevice(device: RokuDeviceDetails): Promise<boolean> {
+        // Increment and capture sequence number to handle concurrent refresh calls
+        const currentSeq = (this.resolveDeviceSequence.get(device.id) ?? 0) + 1;
+        this.resolveDeviceSequence.set(device.id, currentSeq);
+
+        // Set to pending during health check with immediate UI feedback
+        const existingDevice = this.devices.find(d => d.id === device.id);
```
TwitchBronBron, line 277: There's a risk that multiple devices have the same deviceID in this list (rare but not impossible). So make sure we're handling that situation. For now maybe:
- just loop over every device with this deviceId
- or avoid adding duplicate deviceIds
- or when we resolve a device, destroy the cached data on any other devices with that same devcieId (because they're probably lying or wrong)

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,567 @@
...
+
+    private notifyFocusLost() {
+        this.networkChangeMonitor.stop();
+        this.finder.onFocusLost();
+    }
+
+    /**
+     * Add or update a device in the devices array
+     */
+    private upsertDevice(device: RokuDeviceDetails): void {
```
TwitchBronBron: I keep trying to like this, but I just keep wishing it was called `setDevice`. 😆 😢

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -0,0 +1,567 @@
...
+
+export interface RokuDeviceDetails {
+    location: string;
+    id: string;
+    ip: string;
+    deviceState: DeviceState;
+    deviceInfo: DeviceInfoRaw;
+}
+
+function throttleBounce<T extends (...args: any[]) => void>(
```
TwitchBronBron, line 647: could use a comment explaining what this does

PR #703, opened 2026-03-12
Fix bug showing too many "new device" notifications

chrisdp:
fixes an issue where too many notifications where shown via ssdp events. Also, now dismiss this notification after 2 seconds

### DIFF
--- a/src/deviceDiscovery/DeviceManager.ts
+++ b/src/deviceDiscovery/DeviceManager.ts
@@ -445,7 +445,6 @@ export class DeviceManager {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


src/deviceDiscovery/DeviceManager.ts
```diff
@@ -444,11 +445,11 @@ export class DeviceManager {
                 return;
             }
 
+            await this.randomDelay(400, 1_000);
```
TwitchBronBron: Gonna delete this entirely for now. it's making the device picker popup feel too squishy. We can implement UI behavior in a future PR
```suggestion
```

PR #709, opened 2026-03-13
Support hand curated device lists

Christian-Holbrook:
Allows developers to configure roku devices by IP address. This allows developers to always see the devices they have configured.

### DIFF
--- a/src/BrightScriptCommands.ts
+++ b/src/BrightScriptCommands.ts
@@ -466,13 +466,13 @@ export class BrightScriptCommands {
 
             const newDevice = {
                 host: device.ip,
-                name: device.deviceInfo['user-device-name'] || device.deviceInfo['default-device-name'] || device.ip,
                 ...(device.serialNumber && { serialNumber: device.serialNumber })
             };
             userDevices.push(newDevice);
 
             await config.update('devices', userDevices, vscode.ConfigurationTarget.Global);
-            void vscode.window.showInformationMessage(`Added "${newDevice.name}" to user settings.`);
+            const displayName = device.deviceInfo['user-device-name'] || device.deviceInfo['default-device-name'] || device.ip;
+            void vscode.window.showInformationMessage(`Added "${displayName}" to user settings.`);
         });
 
         this.registerCommand('addDeviceToWorkspaceSettings', async (deviceOrItem: { key: string }) => {
@@ -493,13 +493,13 @@ export class BrightScriptCommands {
 
             const newDevice = {
                 host: device.ip,
-                name: device.deviceInfo['user-device-name'] || device.deviceInfo['default-device-name'] || device.ip,
                 ...(device.serialNumber && { serialNumber: device.serialNumber })
             };
             workspaceDevices.push(newDevice);
 
             await config.update('devices', workspaceDevices, vscode.ConfigurationTarget.Workspace);
-            void vscode.window.showInformationMessage(`Added "${newDevice.name}" to workspace settings.`);
+            const displayName = device.deviceInfo['user-device-name'] || device.deviceInfo['default-device-name'] || device.ip;
+            void vscode.window.showInformationMessage(`Added "${displayName}" to workspace settings.`);
         });
 
         this.registerCommand('setDevicePassword', async (deviceIp: string) => {


src/deviceDiscovery/DeviceManager.ts
```diff
@@ -130,6 +129,12 @@ export class DeviceManager {
             if (event?.affectsConfiguration('brightscript.deviceDiscovery.concealDeviceInfo')) {
                 this.emitDevicesChanged();
             }
+
+            //if the `devices` setting was changed, re-apply configured devices and health check them
+            if (event?.affectsConfiguration('brightscript.deviceDiscovery.devices')) {
+                this.loadConfiguredDevices();
+                void this.checkDevicesHealth(true);
```
TwitchBronBron: Don't think we need to health-check these devices until they're shown in the UI.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -558,30 +585,140 @@ export class DeviceManager {
...
+            'user-device-name': configured.name || configured.host,
+            'default-device-name': configured.name || 'Configured Device',
+            'device-id': configured.deviceId || configured.host,
+            'model-number': '',
+            'model-name': '',
+            'software-version': '',
+            'is-tv': 'false',
+            'is-stick': 'false',
+            'developer-enabled': 'true'
+        } as DeviceInfoRaw;
```
TwitchBronBron: let's only keep actual data we know. (so ip address, and serial number?)

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -627,24 +764,52 @@ export class DeviceManager {
...
         const isNewDevice = index < 0;
 
         if (isNewDevice) {
-            this.devices.push(device);
+            this.devices.push({
+                ...device,
+                deviceInfo: {
+                    ...device.deviceInfo,
+                    // If configured, use configured name over discovered name
+                    'user-device-name': device.configuredDevice?.name || device.deviceInfo?.['user-device-name']
```
TwitchBronBron: Let's _not_ overwrite any device-info values. Instead, we can update our UI logic to read from `configuredDevice.name` first, and fall back to `user-device-name`.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -627,24 +764,52 @@ export class DeviceManager {
...
+            const existing = this.devices[index];
+            const configuredDevice = device.configuredDevice ?? existing.configuredDevice;
+            this.devices[index] = {
+                ...existing,
+                ...device,
+                configuredDevice: configuredDevice,
+                // If configured, use configured name over discovered name
+                deviceInfo: {
+                    ...device.deviceInfo,
+                    'user-device-name': configuredDevice?.name || device.deviceInfo?.['user-device-name'] || existing.deviceInfo?.['user-device-name']
```
TwitchBronBron: Same as above, don't overwrite device-info values

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -692,12 +880,27 @@ export class DeviceManager {
 
 export type DeviceState = 'offline' | 'pending' | 'online';
 
+/**
+ * User-configured device from settings (brightscript.deviceDiscovery.devices)
+ */
+export interface ConfiguredDevice {
+    host: string;
+    name?: string;
+    deviceId?: string;
```
TwitchBronBron: Should be `serialNumber` based on recent discussions (once that other `serialNumber` changees land)

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -692,12 +880,27 @@ export class DeviceManager {
...
     location: string;
     id: string;
     ip: string;
     deviceState: DeviceState;
     deviceInfo: DeviceInfoRaw;
+    /**
+     * If present, this device was configured by the user.
+     * Configured devices are never auto-removed.
+     */
+    configuredDevice?: ConfiguredDevice;
```
TwitchBronBron: Let's merge these properties on to `RokuDeviceDetails` itself, I don't think it's overly necessary to keep a separate copy since we can always rebuild it from the configs anyway.

src/BrightScriptCommands.ts
```diff
@@ -434,6 +434,84 @@ export class BrightScriptCommands {
             }
         });
 
+        this.registerCommand('clearConfiguredDevices', async () => {
```
Christian-Holbrook: @TwitchBronBron  I think you don't want this to exist right?

TwitchBronBron: Correct. Nix it.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -197,14 +205,22 @@ export class DeviceManager {
...
      */
     public getAllDevices(): RokuDeviceDetails[] {
         this.firstRequestForDevices = false;
         return [...this.devices].sort(
+            // Configured devices first
             firstBy<RokuDeviceDetails>((a, b) => {
+                if (a.isConfigured !== b.isConfigured) {
+                    return a.isConfigured ? -1 : 1;
+                }
+                return 0;
```
TwitchBronBron: Let's not sort by configured. That can be a future enhancement when we split them out by various categories.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -564,30 +591,148 @@ export class DeviceManager {
...
+        const configuredDevices = Array.from(deviceMap.values());
+        const configuredKeys = new Set(configuredDevices.map(c => c.serialNumber || c.host));
+
+        // Remove (or mark unconfigured) any devices from the devices list that are no longer configured
+        for (let i = this.devices.length - 1; i >= 0; i--) {
+            const device = this.devices[i];
+
+            //skip non-configured devices for this loop
+            if (!device.isConfigured) {
+                continue;
```
TwitchBronBron, line 696: This logic is a little weird, what about devices that haven't had their deviceInfo fetched yet?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -564,30 +591,148 @@ export class DeviceManager {
...
+                configuredPassword: configured.password
+            });
+        }
+    }
+
+    /**
+     * Create placeholder device info for configured devices that haven't been resolved yet.
+     * Only includes data we actually know - everything else will be populated when resolved.
+     */
+    private createPlaceholderDeviceInfo(configured: ConfiguredDevice): DeviceInfoRaw {
```
TwitchBronBron: Not sure we need to do this? We can store `serialNumber` on the device object itself (as seen in `RokuDeviceDetails`) to check whether configured devices match.

src/GlobalStateManager.ts
```diff
@@ -16,7 +16,8 @@ export class GlobalStateManager {
         debugProtocolPopupSnoozeUntilDate: 'debugProtocolPopupSnoozeUntilDate',
         debugProtocolPopupSnoozeValue: 'debugProtocolPopupSnoozeValue',
         lastSeenDevicesByNetwork: 'lastSeenDevicesByNetwork',
-        deviceCache: 'deviceCache'
+        deviceCache: 'deviceCache',
+        ipToSerialNumber: 'ipToSerialNumber'
```
TwitchBronBron:
```suggestion
        ipToSerialNumber: 'serialNumberByIpForNetwork'
```

src/GlobalStateManager.ts
```diff
@@ -162,6 +163,111 @@ export class GlobalStateManager {
...
+     * Clear the IP→serialNumber map
+     */
+    public clearIpToSerialNumberMap(): void {
+        void this.context.globalState.update(this.keys.ipToSerialNumber, undefined);
+    }
+
+    /**
+     * Clear expired entries from the IP→serialNumber map (same expiration as other cached data)
+     */
+    public clearExpiredIpMappings(): void {
```
TwitchBronBron: Clean this up every 24h. you know how, cuz we talked about it

package.json
```diff
@@ -2177,6 +2187,34 @@
                         "default": false,
                         "description": "If set to true, the extension will randomize the numbers and letters in the following fields (useful for hiding your sensitive device fields when creating public screenshots or demos). ['udn', 'device-id', 'advertising-id', 'wifi-mac', 'ethernet-mac', 'serial-number', 'keyed-developer-id']",
                         "scope": "resource"
+                    },
+                    "brightscript.deviceDiscovery.devices": {
```
TwitchBronBron: This should be called `brightscript.devices`

TwitchBronBron (approved):

PR #710, opened 2026-03-13
Always show alive broadcasts

Christian-Holbrook:
We were hiding devices that broadcast they are alive if they were already in the device list.

There is a use case where a developer might pull the plug and replug in the device waiting for the notification to signal the device is online. With the recent device discovery change, we likely won't know a device is offline if it's only offline for a couple of minutes.

Also worth noting that this notification can be disabled with this config value: `"brightscript.deviceDiscovery.showInfoMessages"` set to false.

### DIFF
--- a/src/deviceDiscovery/DeviceManager.ts
+++ b/src/deviceDiscovery/DeviceManager.ts
@@ -464,7 +464,7 @@ export class DeviceManager {
         // Look up cached device by serial number or IP
         const cachedDevice = this.getDevice(serialNumber) ?? this.devices.find(d => d.ip === ip);
 
-        const displayName = cachedDevice?.deviceInfo?.['default-device-name'] ?? `${ip} (${serialNumber ?? 'unknown'})`;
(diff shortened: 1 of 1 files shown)


src/deviceDiscovery/DeviceManager.ts
```diff
@@ -458,25 +446,36 @@ export class DeviceManager {
...
+     */
+    private handleDeviceOnline(ip: string, serialNumber?: string): void {
+        if (!this.showInfoMessages) {
+            return;
+        }
+
+        // Look up cached device by serial number or IP
+        const cachedDevice = this.getDevice(serialNumber) ?? this.devices.find(d => d.ip === ip);
+
+        const displayName = cachedDevice?.deviceInfo?.['default-device-name'] ?? `${ip} (${serialNumber ?? 'unknown'})`;
```
TwitchBronBron: I think I'd rather avoid showing `unknown`, and instead just leave that part blank if not available?

PR #714, opened 2026-03-19
Migrate to serial-number instead of deviceId

Christian-Holbrook:
We are going to use  `serial-number` instead of `deviceId` because the SSDP alive notification gives us the `serial-number` which we can utilze before making a deviceInfo request

### DIFF
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -60,14 +60,6 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
         showHiddenVariables: false,
         enableDebuggerAutoRecovery: false,
         stopDebuggerOnAppExit: false,
-        profiling: {
-            tracing: {
-                enable: true,
-                dir: '${workspaceFolder}/profiling',
-                filename: '${appTitle}_${timestamp}.perfetto-trace',
-                connectOnStart: false
-            }
-        },
         autoRunSgDebugCommands: [],
         files: [...DefaultFiles],
         enableSourceMaps: true,
--- a/src/editors/PerfettoEditor.ts
+++ b/src/editors/PerfettoEditor.ts
@@ -62,8 +62,7 @@ export class PerfettoEditorProvider implements vscode.CustomReadonlyEditorProvid
                             margin: 0;
                             padding: 0;
                             overflow: hidden;
-                            background-color: #ffffff;
-                            color: #000000;
+                            background-color: var(--vscode-editor-background);
                         }
                         iframe {
                             border: none;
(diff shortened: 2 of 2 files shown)


src/deviceDiscovery/DeviceManager.ts
```diff
@@ -175,7 +175,7 @@ export class DeviceManager {
             this.systemSleepMonitor.start();
 
             this.activateMonitoring().then(() => {
-                const lastSeenDeviceIds = this.globalStateManager.getLastSeenDeviceIds(this.networkId);
+                const lastSeenDeviceIds = this.globalStateManager.getLastSeenIds(this.networkId);
```
TwitchBronBron: `getLastSeenDevices`

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -230,10 +230,10 @@ export class DeviceManager {
     }
 
     /**
-     * Get a device by its ID
+     * Get a device by its ID (serial number)
      */
-    public getDeviceById(deviceId: string): RokuDeviceDetails | undefined {
-        return this.devices.find(d => d.id === deviceId);
+    public getDeviceById(id: string): RokuDeviceDetails | undefined {
```
TwitchBronBron:
```suggestion
    public getDevice(serialNumber: string): RokuDeviceDetails | undefined {
```

src/GlobalStateManager.ts
TwitchBronBron:
```suggestion
    serialNumber: string;
```

src/viewProviders/DevicesViewProvider.ts
```diff
@@ -416,8 +416,8 @@ class DeviceDecorationProvider implements vscode.FileDecorationProvider {
             return undefined;
         }
 
-        const deviceId = uri.path.slice(1); // Remove leading slash
-        const state = this.deviceStates.get(deviceId);
+        const id = uri.path.slice(1); // Remove leading slash
+        const state = this.deviceStates.get(id);
```
TwitchBronBron:
```suggestion
        const serialNumber = uri.path.slice(1); // Remove leading slash
        const state = this.deviceStates.get(serialNumber );
```

src/BrightScriptCommands.ts
```diff
@@ -414,7 +414,7 @@ export class BrightScriptCommands {
         this.registerCommand('setActiveDevice', async (deviceOrItem: string | { key: string }) => {
             let ip: string;
             if (typeof deviceOrItem === 'object' && deviceOrItem?.key) {
-                ip = this.deviceManager.getDeviceById(deviceOrItem.key)?.ip;
+                ip = this.deviceManager.getDevice(deviceOrItem.key)?.ip;
```
TwitchBronBron:
```suggestion
                const serialNumber = deviceOrItem.key
                ip = this.deviceManager.getDevice(serialNumber)?.ip;
```

PR #747, opened 2026-04-02
Configured device bug fixes

Christian-Holbrook:
Preserve existing values only when the property is omitted from input, not when explicitly set to undefined. This allows config loading to clear values (e.g., removing a device name) while discovery still preserves config it doesn't know about.                                                            
Resolve hostnames to IPs before looking up serials, since the IP→serial mapping requires resolved IPs. And use these resolved IPs to check if a device is still configured.

### DIFF
--- a/src/deviceDiscovery/DeviceManager.ts
+++ b/src/deviceDiscovery/DeviceManager.ts
@@ -258,13 +258,16 @@ export class DeviceManager {
             let discovered: DiscoveredDeviceEntry | undefined;
 
             if (configured.serialNumber) {
+                // Config has serial - ONLY match by serial (serial is primary key)
                 discoveredIdx = this.discoveredDevices.findIndex(d => d.serialNumber === configured.serialNumber);
-            }
-            if (discoveredIdx < 0 && configured.resolvedIp) {
-                discoveredIdx = this.discoveredDevices.findIndex(d => d.ip === configured.resolvedIp);
-            }
-            if (discoveredIdx < 0) {
-                discoveredIdx = this.discoveredDevices.findIndex(d => d.ip === configured.host);
+            } else {
+                // Config has no serial - match by IP
+                if (configured.resolvedIp) {
+                    discoveredIdx = this.discoveredDevices.findIndex(d => d.ip === configured.resolvedIp);
+                }
+                if (discoveredIdx < 0) {
+                    discoveredIdx = this.discoveredDevices.findIndex(d => d.ip === configured.host);
+                }
             }
 
             if (discoveredIdx >= 0) {
@@ -287,12 +290,14 @@ export class DeviceManager {
             const discovered = this.discoveredDevices[i];
             const device = this.buildMergedDevice(undefined, discovered);
             if (device) {
-                // Check for duplicate by key or IP
+                // Check for duplicate by key
                 if (mergedDevices.has(device.key)) {
                     continue;
                 }
+                // Only skip by IP if neither device has a serial (serial is primary key)
+                // Different serials at same IP = different devices
                 const existingByIp = Array.from(mergedDevices.values()).find(d => d.ip === device.ip);
-                if (existingByIp) {
+                if (existingByIp && !device.serialNumber && !existingByIp.serialNumber) {
                     continue;
                 }
                 mergedDevices.set(device.key, device);
@@ -852,19 +857,30 @@ export class DeviceManager {
 
     /**
      * Update resolvedIp for a configured device by IP or serial number.
+     * Also marks configured devices as offline when a different device is found at their IP.
      */
     private updateConfiguredDeviceResolvedIp(ip: string, serialNumber: string | undefined): void {
-        // Try to find by serial first (more reliable), then by IP
-        let entry: ConfiguredDeviceEntry | undefined;
+        // Try to find by serial first (more reliable)
         if (serialNumber) {
-            entry = this.getConfiguredDevice({ serialNumber: serialNumber });
-        }
-        if (!entry) {
-            entry = this.getConfiguredDevice({ ip: ip });
+            const entryBySerial = this.getConfiguredDevice({ serialNumber: serialNumber });
+            if (entryBySerial) {
+                // Found configured device with matching serial - update its resolved IP
+                entryBySerial.resolvedIp = ip;
+                return;
+            }
         }
 
-        if (entry) {
-            entry.resolvedIp = ip;
+        // No configured device with this serial - check if there's one configured at this IP
+        const entryByIp = this.getConfiguredDevice({ ip: ip });
+        if (entryByIp) {
+            if (entryByIp.serialNumber && serialNumber && entryByIp.serialNumber !== serialNumber) {
+                // Configured device has a different serial than what's actually at this IP
+                // Mark the configured device as offline (it's not where it should be)
+                this.setDeviceState({ serialNumber: entryByIp.serialNumber }, 'offline');
+            } else if (!entryByIp.serialNumber) {
+                // Configured device has no serial - update its resolvedIp via IP matching
+                entryByIp.resolvedIp = ip;
+            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/deviceDiscovery/DeviceManager.ts
```diff
@@ -180,78 +190,114 @@ export class DeviceManager {
...
      * Get device by IP or serial number.
      * Returns device with deviceInfo hydrated from cache.
+     * Respects includeNonDeveloperDevices setting.
      *
      * @param lookup - Object with optional ip and/or serialNumber
-     * @returns Device with deviceInfo or undefined if not found
+     * @returns Device with deviceInfo or undefined if not found/filtered
      */
     public getDevice(lookup: { ip?: string; serialNumber?: string }): RokuDevice | undefined;
     public getDevice(keyOrLookup: string | { ip?: string; serialNumber?: string }): RokuDevice | undefined {
```
TwitchBronBron, line 205: This function should return a device without any type of filter. If we know anything about it, we return it.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -313,15 +376,25 @@ export class DeviceManager {
...
     public clearCurrentDeviceList() {
-        // Keep configured devices, remove discovered-only devices
-        this.devices = this.devices.filter(d => d.isConfigured);
+        // Clear discovered devices (ephemeral)
+        this.discoveredDevices = [];
+
+        // Reset configured devices to pending (they persist but need re-verification)
+        for (const entry of this.configuredDevices) {
+            entry.deviceState = 'pending';
+        }
```
TwitchBronBron: maybe delete this? not sure we need to mark the configured devices as pending.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -964,50 +890,211 @@ export class DeviceManager {
...
 
-            const config: any = util.getConfiguration('brightscript') || {};
-            const includeNonDeveloperDevices = config?.deviceDiscovery?.includeNonDeveloperDevices === true;
-            const developerEnabled = deviceInfo['developer-enabled'] === 'true';
+        // IP dedupe: find existing entry at same IP
+        const existingIdx = this.discoveredDevices.findIndex(d => d.ip === ip);
+        if (existingIdx >= 0) {
+            const existing = this.discoveredDevices[existingIdx];
+            // Don't downgrade from online to pending
+            const newState = (existing.deviceState === 'online' && deviceState === 'pending')
```
TwitchBronBron: should you call computeMergedState instead?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -964,50 +890,211 @@ export class DeviceManager {
...
-     * Process a discovered IP address from SSDP.
-     * Fetches device info, applies filtering, and sets if valid.
-     * @param serialNumber - Serial number from SSDP USN header, if available
+     * Add or update a device in the discoveredDevices array.
+     * Handles deduplication by serial number (removes old IP entry if serial matches).
      */
-    private async processDiscoveredIp(ip: string, serialNumber?: string): Promise<void> {
-        try {
-            const deviceInfo = await this.fetchDeviceInfo(ip, 8060);
+    private setDiscoveredDevice(ip: string, serialNumber: string | undefined, deviceState: 'pending' | 'online'): void {
```
TwitchBronBron: Let's eliminate the deviceState param, we can determine "freshness" by looking at the cache for deviceInfo. And then also keep "online" sticky if it's set.

src/deviceDiscovery/DeviceManager.ts
TwitchBronBron, line 205: let's make this one alwyas return a device, no matter what. (no filtering)

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -180,78 +190,114 @@ export class DeviceManager {
...
     }
 
     /**
-     * Get a list of all roku devices known by this extension (by scanning, hardcoded lists, etc)
-     * Returns full devices with deviceInfo hydrated from cache.
-     * Configured devices are sorted first, then by form factor, name, and id.
+     * Get a list of all visible roku devices.
+     * Respects includeNonDeveloperDevices setting.
      */
     public getAllDevices(): RokuDevice[] {
```
TwitchBronBron, line 236: should have a public function that is "get every device in the manager, no filtering". That one probably should be called getAllDevices()? Then there can be another function for "getFilteredDevices" or "getDevicesForUI", idk the name...that represents the list that should be displayed to the users in UI.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -275,6 +321,23 @@ export class DeviceManager {
         );
     }
 
+    /**
+     * Compute merged device state from configured and discovered states.
+     * Priority: online > offline > pending
+     */
+    private computeMergedState(
```
TwitchBronBron: Should use this function any place we're doing the comparison (i think i saw inline comparions a few spots). Also, not the best name for the function. maybe 'computeState' or 'computeDeviceState' rather than 'merged'?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -646,145 +578,111 @@ export class DeviceManager {
      */
     private async loadConfiguredDevices(): Promise<void> {
```
TwitchBronBron, line 661: Should probably sort each of these device lists, that way their merging applies deterministically.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -646,145 +578,111 @@ export class DeviceManager {
...
-                ip: ip,
-                serialNumber: serialNumber,
-                deviceState: deviceState,
-                isConfigured: true,
-                configuredIn: configured.configuredIn,
-                isDiscovered: isDiscovered,
-                configuredName: configured.name,
-                configuredPassword: configured.password
-            });
         }
```
TwitchBronBron, line 723: should probably sort the final result before exiting this function, so the devices are more deterministically processed in other places.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -814,31 +712,51 @@ export class DeviceManager {
...
     }
 
-    private async checkDevicesHealth(force = false): Promise<void> {
-        // Use internal devices array directly - no need to hydrate from cache
-        const devices = this.devices;
+    /**
+     * Update state for a configured device by IP or serial number.
+     */
+    private updateConfiguredDeviceState(ip: string, serialNumber: string | undefined, state: DeviceState): void {
```
TwitchBronBron: Any way to merge this with the other "compute device state" function? A little confusing which one to use, which one does what, etc. Might not make sense....idk

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -814,31 +712,51 @@ export class DeviceManager {
...
+        if (entry) {
+            entry.deviceState = state;
+            // Update resolvedIp if we have a successful resolution
+            if (state === 'online' && ip) {
+                entry.resolvedIp = ip;
+            }
+        }
+    }
+
+    private async checkDevicesHealth(force = false, doSyntheticDelay = true): Promise<void> {
```
TwitchBronBron: this name is way too similar to `checkDeviceHealth`. I wonder if we could name them:
- `healthCheckDevice`
- `healthCheckAllDevices`

? Thoughts?

TwitchBronBron (changes requested):

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -964,50 +890,211 @@ export class DeviceManager {
...
+    private getConfiguredDeviceBySerial(serialNumber: string): ConfiguredDeviceEntry | undefined {
+        return this.configuredDevices.find(d => d.serialNumber === serialNumber);
+    }
+
+    /**
+     * Find a configured device by host or resolved IP.
+     */
+    private getConfiguredDeviceByIp(ip: string): ConfiguredDeviceEntry | undefined {
+        return this.configuredDevices.find(d => d.resolvedIp === ip || d.host === ip);
+    }
```
TwitchBronBron: could we merge these into a `getConfiguredDevice` function that takes options with the 2 supported props? `{ serialNumber: string} | { ip: string}`

TwitchBronBron (approved):

PR #761, opened 2026-04-15
Auto-discover Roku projects and generate debug configs

chrisdp:
## Summary

- Adds `RokuProjectManager` which watches the workspace for `bsconfig*.json` and `brsconfig*.json` files, registers build tasks for discovered projects, and surfaces them as debug configurations
- `BsConfigProjectProvider` handles bsconfig-based projects: runs `npx bsc` as a pre-launch task and uses the staging dir as the debug root
- `BrsConfigProjectProvider` handles the new `brsconfig*.json` format: no build step required, the `files` and `rootDir` from the config are passed directly to the debug session
- When F5 is pressed with no `launch.json`, the extension resolves a debug config from the active file by finding which project config owns it — if multiple configs claim the file, a quick-pick lets the user choose
- New UI: a "Debug Roku Project" status bar button appears whenever projects are discovered, and a CodeLens on each config file for one-click debugging


src/managers/RokuProject/RokuProjectManager.ts
```diff
@@ -0,0 +1,371 @@
...
+     * Called after a config URI is registered. Providers may use this to update
+     * internal file-ownership indexes.
+     */
+    afterConfigRegistered?(configUri: vscode.Uri): void;
+    /**
+     * Called after a config URI is unregistered. Providers may use this to clean
+     * up entries in internal file-ownership indexes.
+     */
+    afterConfigUnregistered?(configUri: vscode.Uri): void;
+}
```
TwitchBronBron: Let's move these interfaces to the bottom. I typically prefer having the main class for the file as close to the top as possible.

Would also be useful to have a doc block above DiscoveredRokuProject and DiscoveredRokuProject to explain what they're used for. DiscoveredRokuProject is fairly self-explanatory, but ProjectBuildResult...there are a few ways I can interpret what that's used for from the name so a comment would be helpful

TwitchBronBron: Can we move these interfaces to the bottom? I typically prefer having the main class for the file as close to the top as possible.

Would also be useful to have a doc block above them to explain what they're used for. DiscoveredRokuProject is fairly self-explanatory, but ProjectBuildResult...there are a few ways I can interpret what that's used for from the name so a comment would be helpful

src/managers/RokuProject/RokuProjectManager.ts
```diff
@@ -0,0 +1,371 @@
...
+        statusBarItem.command = DEBUG_ROKU_PROJECT_COMMAND;
+        statusBarItem.text = '$(debug-start) Debug Roku Project';
+        statusBarItem.tooltip = 'Debug Roku Project';
+
+        context.subscriptions.push(
+            statusBarItem,
+            vscode.commands.registerCommand(DEBUG_ROKU_PROJECT_COMMAND, (arg?: vscode.Uri | ProjectTreeItem) => {
+                const uri = arg instanceof ProjectTreeItem ? arg.project.configUri : arg;
+                void this.debugProject(uri);
+            }),
```
TwitchBronBron: We've been slowly migrating our command registrations over into the `commands` folder in some fashion. Thoughts on moving this one over there too? I don't mind having the `debugProject` be public (which would be required if we moved the command), it's fairly straightforward.

src/managers/RokuProject/RokuProjectManager.ts
```diff
@@ -0,0 +1,371 @@
...
+    private readonly providers: ProjectConfigProvider[] = [
+        new BsConfigProjectProvider(),
+        new BrsConfigProjectProvider()
+    ];
+
+    public register(context: vscode.ExtensionContext) {
+        const statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 0);
+        statusBarItem.command = DEBUG_ROKU_PROJECT_COMMAND;
+        statusBarItem.text = '$(debug-start) Debug Roku Project';
+        statusBarItem.tooltip = 'Debug Roku Project';
```
TwitchBronBron: Would be nice to consolidate some of the statusbar logic. Maybe like a single function that just "makes if not made", "adds if not added", then "set all props on it". Then everywhere we can just call .setStatusBarStuff() everywhere in here and it's all centralized instead of spread out.

src/managers/RokuProject/RokuProjectManager.ts
```diff
@@ -0,0 +1,371 @@
...
+        // all workspace folders automatically — no need to iterate them explicitly here.
+        for (const provider of this.providers) {
+            const uris = await provider.findProjectConfigs();
+            for (const uri of uris) {
+                this.registerProject(uri);
+            }
+        }
+    }
+
+    public registerProject(uri: vscode.Uri) {
```
TwitchBronBron: Does this need to be public? If just for tests, you can call `projectManager['registerProject']` there. Might be worth doing a whole pass at what functions we actually intend for external interaction, and move everything else to private.

src/managers/RokuProject/RokuProjectManager.ts
```diff
@@ -0,0 +1,371 @@
...
+
+        // Populate the task registry with whatever is currently in the workspace
+        this.syncProjects().catch(err => {
+            console.error('Error syncing Roku projects:', err);
+        });
+    }
+
+    // -------------------------------------------------------------------------
+    // Status bar
+    // -------------------------------------------------------------------------
```
TwitchBronBron: Really loving these random comment blocks. they're not our normal style, and they're also not `#region` style comments either. Are they necessary? If not, maybe we can remove them since the method and prop names are already fairly clear what each thing is for.

Also there seems to only be like 1-2 functions "belonging" to each block, which isn't much different than just having an actual docblock comment above a function instead...(which I'd much prefer)

src/managers/RokuProject/BsConfigProjectProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+        } catch {
+            // bsconfig is invalid — skip silently
+        }
+    }
+
+    public afterConfigUnregistered(configUri: vscode.Uri): void {
+        this.configIndex.delete(configUri.fsPath);
+    }
+
+    public buildProject(configUri: vscode.Uri): ProjectBuildResult {
```
TwitchBronBron: this function name is deceiving. I don't think it actuall _builds_ anything? It returns configurations that can be _used_ to build. 

I'd rather see this named more on-the-nose, something like `assembleTaskAndLaunchConfig` or `assembleBuildInfo`.

src/managers/RokuProject/BrsConfigProjectProvider.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    public readonly configFileSelector: vscode.DocumentFilter[] = [
+        { pattern: '**/brsconfig*.json', scheme: 'file' }
+    ];
+
+    /**
+     * Per-config store keyed by configUri.fsPath.
+     * Populated by afterConfigRegistered; used by findProjectConfigFromFile.
+     */
+    private readonly configIndex = new Map<string, IndexedConfig>();
```
TwitchBronBron: this name `index` is throwing me off a bit. It does't really have anything to do with an index. I think this is something more like `configByPath` or `configLookup`?

src/managers/RokuProject/BsConfigProjectProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    public readonly configFileSelector: vscode.DocumentFilter[] = [
+        { pattern: '**/bsconfig*.json', scheme: 'file' }
+    ];
+
+    /**
+     * Per-config store keyed by configUri.fsPath.
+     * Populated by afterConfigRegistered; used by findProjectConfigFromFile.
+     */
+    private readonly configIndex = new Map<string, IndexedConfig>();
```
TwitchBronBron: Same comment as from BrsConfigProvider. `index` is weird in this name

src/managers/RokuProject/BsConfigProjectProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+        const flavor = configFilename.replace(/^bsconfig\W?(.*)\.json$/, '$1').replace(/\W+/g, ' ').trim();
+        const taskName = `build ${vscode.workspace.asRelativePath(configUri)}`;
+
+        const taskConfig: TaskConfig = {
+            // --project: explicit so bsc doesn't have to search for the config file
+            // --retain-staging-folder: keep the staging dir after the build so the debugger can use it as rootDir
+            // --source-map: generate source maps for debugging
+            // --copy-to-staging: copy files to the staging dir even if they don't need to be transformed by bsc, so that the staging dir is a complete set of the files needed for debugging
+            // --create-package false: don't create a zip package since the staging dir is used directly as the debug root
+            command: `npx bsc --project "${configFilename}" --retain-staging-folder --source-map --copy-to-staging --create-package false`,
```
TwitchBronBron: Do we want to help the user out at all with "we couldn't find brighterscript anywhere in your path so it's probably not gonna work" prior to running the task? For devs who know "oh yeah, i'm supposed to install brighterscript", the normal behavior would be fine. But for people testing out sample projects and stuff, it might be nicer to guide them slightly. like "You tried to run a brighterscript project but you likely forgot to run `npm install` (because we couldn't find npx bsc anywhere in your local path". ? Just a thought.

package.json
```diff
@@ -188,6 +189,12 @@
         },
         "views": {
             "vscode-brightscript-language": [
+                {
+                    "id": "rokuProjectsView",
+                    "name": "Roku Projects",
+                    "icon": "$(project)",
+                    "when": "brightscript.hasRokuProjects && false"
```
TwitchBronBron, line 197: Yeah, I agree with hiding the projects view for now. We still have a lot to discuss around how to handle "runnable projects" vs "language server projects".

src/extension.ts
```diff
@@ -160,9 +171,12 @@ export class Extension {
...
         //register the debug configuration provider
-        let configProvider = new BrightScriptDebugConfigurationProvider(context, this.telemetryManager, this.extensionOutputChannel, userInputManager, this.brightScriptCommands);
+        let configProvider = new BrightScriptDebugConfigurationProvider(context, this.telemetryManager, this.extensionOutputChannel, userInputManager, this.brightScriptCommands, rokuProjectProvider);
         context.subscriptions.push(
-            vscode.debug.registerDebugConfigurationProvider('brightscript', configProvider)
+            // Initial: resolveDebugConfiguration — handles launch.json configs and F5 with no launch.json
+            vscode.debug.registerDebugConfigurationProvider('brightscript', configProvider, vscode.DebugConfigurationProviderTriggerKind.Initial),
+            // Dynamic: provideDebugConfigurations — surfaces discovered Roku projects in the debug picker
+            vscode.debug.registerDebugConfigurationProvider('brightscript', configProvider, vscode.DebugConfigurationProviderTriggerKind.Dynamic)
         );
```
TwitchBronBron, line 182: Is this necessary? Just curious, because I would sort-of expect the normal `registerDebugConfigurationProvider` to just assume all trigger kinds? Do we have to be explicit here with the register-for-each-trigger-kind?

src/DebugConfigurationProvider.ts
```diff
@@ -77,11 +79,24 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
...
+    public provideDebugConfigurations(folder?: WorkspaceFolder, _token?: CancellationToken): vscode.DebugConfiguration[] {
+        return this.rokuProjectDiscovery?.provideDebugConfigurations(folder) ?? [];
+    }
+
     /**
      * Massage a debug configuration just before a debug session is being launched,
      * e.g. add all missing attributes to the debug configuration.
      */
     public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, config: BrightScriptLaunchConfiguration, token?: CancellationToken): Promise<BrightScriptLaunchConfiguration> {
+        // F5 with no launch.json — ask RokuProjectDiscovery to find a config from the active file.
```
TwitchBronBron, line 91: Love seeing this in here, helps centralize all our debug configuration resolution logic. Glad we made this change.

src/BrightScriptTaskProvider.ts
```diff
@@ -18,8 +17,45 @@ export class BrightScriptTaskProvider implements vscode.Disposable {
...
     private taskProvider: vscode.Disposable;
+    private registeredTasks = new Map<string, TaskConfig>();
+
+    public registerTask(name: string, config: TaskConfig) {
+        this.registeredTasks.set(name, config);
+    }
+
+    public unregisterTask(name: string) {
+        this.registeredTasks.delete(name);
+    }
```
TwitchBronBron, line 44: Let's add some docblock comments explaining that these functions aren't actually registering the tasks directly with vscode, but rather keeping them around in memory for "when vscode asks for them". 

Also, could be useful to explain what people on the outside typically leverage these functions? (just for more clarity when editing this file).

TwitchBronBron (changes requested):
This looks great! I have a few naming suggestions, and some documentation stuff, but overall I'm good with this implementation.

src/extension.ts
chrisdp: Yes, the default is actually just `Initial`. If you want `Dynamic` you need to sub separately. I added the kind to the first one so it was more explicitly clear that we want both.

TwitchBronBron (approved):

PR #763, opened 2026-04-17
Detect device-online via SSDP alive burst

chrisdp:
## Summary

- `RokuFinder` previously emitted `device-online` on every debounced `ssdp:alive`, so a Roku already on the network would re-trigger "Device Online" notifications on every periodic heartbeat.
- Now `device-online` is inferred from a **burst** of alives (≥2 within a 5s window, per-IP). Devices send a tight burst when they first join the network; steady-state heartbeats are spaced minutes apart and never accumulate.
- The burst tracker re-arms once the window goes quiet so a future reboot fires again, and both the burst map and the `found` debounce map are now self-cleaning via `setTimeout` (removing the old 5-minute periodic sweep).
- `found` is emitted before burst detection runs so discovery is never skipped if the tracker throws.
- With burst detection at the source, the redundant 500ms per-device notification debounce in `DeviceManager.handleDeviceOnline` is removed (the `deviceOnlineNotifiers` map and the lodash `debounce` import go with it).

### DIFF
--- a/src/deviceDiscovery/DeviceManager.ts
+++ b/src/deviceDiscovery/DeviceManager.ts
@@ -10,7 +10,6 @@ import { NetworkChangeMonitor, getNetworkHash } from './NetworkChangeMonitor';
 import { SystemSleepMonitor } from './SystemSleepMonitor';
 import { util } from '../util';
 import { vscodeContextManager } from '../managers/VscodeContextManager';
-import { debounce } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


TwitchBronBron (approved):

TwitchBronBron:
This isn't an accurate solution. Roku sends 3 message bursts in both situations (when powering on, and when saying "I'm still here").

PR #772, opened 2026-04-24
Device credentials refactor: SN-keyed passwords and smart resolution

### LINKED ISSUE #687: Remember password from last ${promptForPassword}
Support remembering the password from the last time we showed the ${promptForPassword} popup. This can be stored in global 

Here's the function where it's triggered. 
https://github.com/rokucommunity/vscode-brightscript-language/blob/22c4d8db2085012d5b015e0c152799de01ed0d1a/src/DebugConfigurationProvider.ts#L453-L464

I would expect we would store the last-known-password for a given host, and look it up right before showing the popup. Then save the password by that IP address before returning the value. 

```typescript
const ipAddress = '1.2.3.4';
const password = 'aaaa';
await this.context.workspaceState.update('lastUsedPassword-${ipAddress}', password);
```

and similar `this.context.workspaceState.get('lastUsedPassword-${ipAddress}')` for reading it.

chrisdp:
## Summary

- **SN-keyed credentials.** Passwords are stored per-device by serial number instead of per-IP, so a device's stored password survives DHCP moves, network changes, and restarts. `brightscript.devices[].password` in user / workspace / workspace-folder settings is the top-priority source; the new `CredentialStore` acts as a running cache of validated passwords.
- **Candidate-based password resolution for debug launches.** `DebugConfigurationProvider` now collects every known password (cred store, SN-matched settings entries, `brightscript.defaultDevicePassword`, merged result, raw launch.json), dedupes, and validates each against the device in priority order. The first accepted candidate wins and caches forward. If every candidate is rejected, the user is prompted with an inline Remember toggle. Bad-password re-prompts until the user gives up; unreachable terminates the flow.
- **Manual-IP flows health-check before accepting.** The picker's manual-IP entry and `promptForHostManual` both probe the device via the now-public `DeviceManager.processDiscoveredIp` before resolving, and show "Unable to connect to a Roku at ..." with a retry loop otherwise.
- **Legacy `workspaceState.devicePasswords` migration.** Pre-refactor IP-keyed entries are peeked on launch, validated once, and either migrated forward (ok), dropped (bad-password), or left in place for a future try (unreachable). Reads never fall back to the legacy store.
- **CredentialStore lifecycle.** `setDevicePassword` writes across every matching `brightscript.devices[]` scope plus the cred store. New `clearDevicePassword` command. The Devices view shows Set / Change / Clear child items that live-update via a new `changed` event on `CredentialStore`. 4-character minimum enforced on the Set prompt, matching Roku's own on-device UI.
- **Gate to avoid silent persistence.** `acceptPassword` only writes to the cred store when the SN is already adopted (has a cred-store entry OR appears in `brightscript.devices[]`). The Remember toggle on the prompt is the explicit opt-in path for new devices.
- **rokuDeploy uplift.** Added `validateDeveloperPassword(...)` to rokuDeploy via `HEAD /plugin_install` over HTTP digest using native `fetch` (rokucommunity/roku-deploy#252).

## Remaining follow-ups (not in this PR)

- [ ] Schema `deprecationMessage` on `\${promptForHost}` / `\${activeHost}` / `\${promptForPassword}` / `\${activeHostPassword}` and rewrite of `docs/variable-substitutions.md`.
- [ ] `setDevicePassword` / `clearDevicePassword`: fall back to a device picker when invoked with no SN arg (currently throw).
- [ ] Stop auto-writing `host` in the save handlers. Needs companion edits to `DeviceManager.loadConfiguredDevices` (which currently early-returns on `!device?.host`) and the `host` `required` entry in the schema.
- [ ] Palette visibility for `setDevicePassword` / `clearDevicePassword` via `package.json` `contributes.commands` (today they are tree-only).

## Test plan

- [x] Full unit suite: 812 passing, 5 pre-existing `LanguageServerManager` failures (unrelated brighterscript path resolution).
- [ ] Launch with hardcoded password — config.password candidate wins, password cached.
- [ ] Launch with empty password on a new device — prompt appears with Remember toggle, toggle off = session-only, toggle on = cached.
- [ ] Launch with empty password on an adopted device — no prompt, cred store wins.
- [ ] Launch with stale cached password — bad-password fallthrough to next candidate, then prompt on exhaustion.
- [ ] Launch against offline host — "device unreachable" surfaces, flow terminates.
- [ ] Pre-existing \`workspaceState.devicePasswords\` entry migrates on first launch (ok or bad paths) and stays put on unreachable.
- [ ] Device picker manual-IP entry against a non-Roku IP — shows "Unable to connect" and keeps picker open.
- [ ] Settings Sync carries \`brightscript.devices[].password\` across machines but not \`CredentialStore\` contents.

### DIFF
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -22,7 +22,9 @@ import type { DeviceInfo } from 'roku-deploy';
 import type { UserInputManager } from './managers/UserInputManager';
 import type { BrightScriptCommands } from './BrightScriptCommands';
 import type { RokuProjectManager } from './managers/RokuProject/RokuProjectManager';
-import type { DeviceManager } from './deviceDiscovery/DeviceManager';
+import type { DeviceManager, RokuDevice } from './deviceDiscovery/DeviceManager';
+import type { CredentialStore } from './managers/CredentialStore';
+import type { ConfiguredDevice } from './GlobalStateManager';
 
 
 export class BrightScriptDebugConfigurationProvider implements DebugConfigurationProvider {
@@ -34,6 +36,7 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
         private userInputManager: UserInputManager,
         private brightScriptCommands: BrightScriptCommands,
         private deviceManager: DeviceManager,
+        private credentialStore: CredentialStore,
         private rokuProjectDiscovery?: RokuProjectManager
     ) {
         this.context = context;
@@ -111,8 +114,9 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
 
             result = await this.sanitizeConfiguration(result, folder);
             result = await this.processEnvFile(folder, result);
-            result = await this.processHostParameter(result);
-            result = await this.processPasswordParameter(result);
+            const [resultAfterHost, device] = await this.processHostParameter(result);
+            result = resultAfterHost;
+            result = await this.processPasswordParameter(config, result, device);
             result = await this.processDeepLinkUrlParameter(result);
             result = await this.processLogfilePath(folder, result);
             result = this.processDapLogFilePath(folder, result);
@@ -450,9 +454,14 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
      * Validates the host parameter in the config and opens an input ui if set to ${promptForHost}.
      * ${activeHost} is a deprecated alias for ${promptForHost}.
      * Both use the active device when it's set and passes a health check, otherwise fall back to the device picker.
+     *
+     * Returns the updated config alongside the probed `RokuDevice` so downstream
+     * password resolution can look up credentials by serial number without
+     * re-fetching device info. Device is undefined when the resolved host is
+     * unreachable or not a developer-enabled Roku.
      * @param config  current config object
      */
-    private async processHostParameter(config: BrightScriptLaunchConfiguration): Promise<BrightScriptLaunchConfiguration> {
+    private async processHostParameter(config: BrightScriptLaunchConfiguration): Promise<[BrightScriptLaunchConfiguration, RokuDevice | undefined]> {
         const trimmedHost = config.host.trim();
         const needsHostPrompt =
             trimmedHost === '' ||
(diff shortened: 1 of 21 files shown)


src/DebugConfigurationProvider.ts
```diff
@@ -476,40 +485,288 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
             await this.context.workspaceState.update('remoteHost', config.host);
         }
 
-        return config;
+        // Probe the resolved host so downstream password resolution has fresh SN/deviceInfo.
+        // Unreachable or filtered hosts yield no registered device; password resolution handles that.
+        await this.deviceManager.processDiscoveredIp(config.host);
+        const device = this.deviceManager.getDevice({ ip: config.host });
+
+        return [config, device];
```
TwitchBronBron, line 492: Why not just do this on the outside? We're doing it "last" here anyway. Just probe the host outside the call and pass that into the password flow. No need to change this return type?

src/GlobalStateManager.ts
```diff
@@ -331,6 +331,13 @@ interface IpToSerialNumberEntry {
     timestamp: number;
 }
 
+export interface ConfiguredDevice {
```
TwitchBronBron: this interface already exists at the bottom of DeviceManager, so we can delete this one.

TwitchBronBron (approved):
Overall, I think this will work. I've got some thoughts about moving code to better locations (specifically the stuff in DebugConfigurationProvider), but all that stuff is private right now so it's pretty easy to improve in the future as we iterate.

I haven't tested any of this, but I think it's structurally and logically fine, from what I can see

src/DebugConfigurationProvider.ts
```diff
@@ -476,40 +484,288 @@ export class BrightScriptDebugConfigurationProvider implements DebugConfiguratio
...
+                });
+                input.onDidHide(() => {
+                    resolve(undefined);
+                });
+                input.show();
+            });
+        } finally {
+            input.dispose();
+        }
+    }
```
chrisdp: consolidate in the future and consider eyeball button for toggling PW visability

TwitchBronBron (approved):

PR #788, opened 2026-05-14
Force-terminate dev channel on restartDevApplication

chrisdp:
## Summary
- `restartDevApplication` now sends `exit-app/dev/true` instead of two `exit-app/dev` calls. The `/true` variant terminates the channel even when it is suspended in the background via Instant Resume, so the second call is no longer needed.
- Added unit tests covering the success path, the missing-dev-channel path, and the post-launch verification warning.


TwitchBronBron (approved):
tested this locally, and it does indeed terminate an instant resume app and start it from scratch

PR #790, opened 2026-05-14
Add filter dropdown to the Devices view

chrisdp:
## Summary
Adds a `$(filter)` icon to the Devices view title bar that opens a submenu with toggle items for each facet:

- **Dev mode**: Dev Mode Enabled / Dev Mode Disabled
- **Form factor**: TV / Set Top Box / Streaming Stick
- **Connectivity**: Online / Offline
- **Source**: User Defined / Auto Detected

Each row uses ☑/☐ glyphs to reflect state. Strict-AND semantics — every device has to match every enabled facet to appear.

## Notes
- Filter state persists per-workspace via `context.workspaceState`. Only keys that diverge from defaults are written; toggling a facet back to its default removes the key from storage.
- Defaults preserve the previous Devices view behavior: online, dev-enabled, all form factors and sources visible. Dev Mode Disabled and Offline are off by default.
- A device that appears in both settings and the SSDP results counts as User Defined.
- Pending devices stay visible to the Online filter only when they were online on the prior check. On first load an unverified pending device falls under Offline so it doesn't briefly flash in the list.
- Independent of the existing `brightscript.deviceDiscovery.includeNonDeveloperDevices` setting — that still acts as a hard floor at the DeviceManager layer.

## Implementation
The submenu uses two paired commands per facet (`*` and `*.active`) and `when` clauses on per-facet context keys (`brightscript.devicesView.filter.<facet>`) to swap which one renders. `command.toggled` and `$(check)` codicons in submenu titles are not supported by the extension API, hence the title-swap approach.

DeviceManager records `lastState` on configured and discovered entries before each transition, exposed on `RokuDevice.lastDeviceState` for the connectivity filter to consult.


docs/device-discovery.md
```diff
@@ -195,7 +195,6 @@ All settings are in the `brightscript.deviceDiscovery` namespace and can be set
 |---|---|---|
 | `brightscript.deviceDiscovery.enabled` | `true` | Enables automatic network scanning and passive SSDP listening for Roku devices. Disable this if you always use a fixed IP address or want no network activity from the extension. |
 | `brightscript.deviceDiscovery.showInfoMessages` | `true` | When `true`, a toast notification appears each time a new Roku device is discovered on the network. |
-| `brightscript.deviceDiscovery.includeNonDeveloperDevices` | `false` | When `true`, the extension includes Roku devices that do **not** have Developer Mode enabled. By default only developer-enabled devices are shown. |
```
TwitchBronBron, line 198: We deleted this one because it's replaced by our new filter props

src/viewProviders/DevicesViewProvider.ts
```diff
@@ -15,21 +22,51 @@ let treeItemKeySequence = 0;
...
+/**
+ * Context key prefix VS Code reads to decide which menu entry (checked vs
+ * unchecked) to show for each filter facet in the title-bar submenu.
+ */
+const DEVICES_VIEW_FILTER_CONTEXT_KEY_PREFIX = 'brightscript.devicesView.filter.';
+
+// Re-export the filter shape so downstream callers (extension.ts, tests) can use them
+// without importing from both modules.
+export type { DeviceFilters };
+export { DEVICE_FILTER_KEYS };
```
TwitchBronBron: Don't need to re-export either of these. have the other places just import from the correct place.

TwitchBronBron (approved):

PR #802, opened 2026-05-21
Route extension logs through the BrightScript Extension output channel

chrisdp:
## Summary
- Adds `src/logging.ts`, a shared logger (built on `@rokucommunity/logger`) that writes to the developer console and to the existing `BrightScript Extension` output channel. The channel is already piped into the optional `brightscript.extensionLogfilePath` setting, so file logging works without further wiring.
- Migrates all `console.*` calls across the extension source (~45 call sites in 19 files) to the shared logger, each file tagged with a component-named sub-logger (e.g. `[LanguageServerManager]`) so messages are attributable in the output channel.
- The LSP keeps its own `BrightScript Language Server` output channel — only extension-level logs route through the shared logger.

## Motivation
End-users currently can't see extension logs — the `console.*` calls scattered across the source only surface in the Extension Host devtools. This routes them to the `BrightScript Extension` output channel (and thus the logfile setting) without each call site needing to know about the channel.

## Notes
- Mapping is behavior-preserving: `console.log`→`logger.log`, `warn`→`warn`, `error`→`error`. (In `@rokucommunity/logger`, `info` is *more* verbose than `log`, so `logger.info` would have been silently dropped at the default `log` level.)
- Skipped: the Perfetto editor's inline webview script (runs in the webview, not the extension host) and `webviews/` sources.
- Open question: should there be a `brightscript.logLevel` setting to control verbosity? A few noisy call sites (e.g. `DeviceManager.resolveDevice` tracing) could then be demoted to `debug`/`trace`.


src/logging.ts
TwitchBronBron, line 1: BrighterScript and roku-debug both call this file `logging.ts`, let's standardize on that.

chrisdp: Renamed to `logging.ts` and also exported a bound `createLogger` helper to match the pattern in brighterscript/roku-debug.

TwitchBronBron (approved):

PR #825, opened 2026-06-12
chore Kill orphaned watch processes on shutdown to fix wedged debug launches

chrisdp:
When the Extension Development Host is relaunched, `watch-all` left its `npm` → `tsc`/`vite` subtree running because only the `ts-node` parent was terminated. Those orphaned watchers piled up across launches, fighting over `dist/` and the debug host until a window reload or full restart was needed.

- `watch-all` now spawns each watcher in its own process group and kills the whole subtree on `SIGINT`/`SIGTERM`/`SIGHUP`/exit.
- On startup it cleans up watchers recorded by a previous run that exited ungracefully (via a per-checkout pid file), which breaks the accumulation cycle.
- Added a cross-platform `reset-dev` task to clear stale watchers and dev ext-host windows when a launch still gets stuck.
- Added `scripts/tsconfig.json` so the editor type-checks the `scripts/` dir (clears the `debounce`/`deferred`/down-level-iteration errors); removed a dead local and an unused import that surfaced.

### DIFF
--- a/scripts/watch-all.ts
+++ b/scripts/watch-all.ts
@@ -1,6 +1,9 @@
 import * as fsExtra from 'fs-extra';
 import * as path from 'path';
-import { spawn } from 'child_process';
+import * as os from 'os';
+import * as crypto from 'crypto';
+import { spawn, spawnSync } from 'child_process';
+import type { ChildProcess } from 'child_process';
 import * as debounce from 'debounce';
 import * as chalk from 'chalk';
 import * as dayjs from 'dayjs';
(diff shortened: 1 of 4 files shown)


scripts/watch-all.ts
```diff
@@ -32,7 +35,68 @@ class Logger {
...
-logger.writeLine(`${timestamp()} Starting compilation in watch mode...`);
+const repoRoot = path.resolve(__dirname, '..');
+//scope the pid file to this checkout so parallel clones don't kill each other's watchers
+const pidFile = path.join(
+    os.tmpdir(),
+    `vscode-brightscript-watch-all-${crypto.createHash('md5').update(repoRoot).digest('hex').slice(0, 8)}.json`
+);
+const watchers: ChildProcess[] = [];
+
+function killGroup(pid: number) {
```
TwitchBronBron, line 46: Let's harden this slightly, to ensure we don't just kill any arbitrary process with that id. (maybe store PID _and_ process name) and compare them before doing the kill.

TwitchBronBron (changes requested):
I'm not totally convinced this is the problem, and i'm not loving the processId concept in general. Let's troubleshoot a few times where you run just the task to prove out that it solves the problem.

PR #831, opened 2026-06-25
Show an empty-state for empty Perfetto trace files

chrisdp:
Opening an empty (zero-byte) `.pftrace` / `.perfetto-trace` file showed a confusing "not a recognized format" error from the embedded Perfetto UI.

This detects the empty case up front and shows a friendly empty-state instead, with a button to open the file in the plain text editor.


src/editors/PerfettoEditor.ts
chrisdp: lets handle this case too, update the error view for template error messages and optional button. Also move the read before this call for the html so we can fail faster.

TwitchBronBron (approved):

PR #832, opened 2026-06-25
Add enablePostfix option to component library launch schema

chrisdp:
Adds the per-component-library `enablePostfix` option to the launch configuration schema so it surfaces in launch.json completion.

`enablePostfix` (default `true`) controls whether a component library's `.brs` files are renamed with a `__lib<index>` postfix during staging. Set it to `false` to leave file names untouched for a library that loads files by a fixed name at runtime.

Depends on the debug-adapter support in rokucommunity/roku-debug#378; this should land after that change is released and the roku-debug dependency is bumped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -853,6 +853,11 @@
                                             "description": "Should this component library be installed onto the Roku device during the sideloading process? Component libraries are installed in the order they're found in this array, so it is recommended to keep all `install: true` component libraries at the top of the array. \n\nTo utilize this component library in your application, set `dcl_channel_ids=DCLHelloLib:devlib_dcl_1` (where 1 is the 1-based position in this array)",
                                             "default": false
                                         },
+                                        "enablePostfix": {
+                                            "type": "boolean",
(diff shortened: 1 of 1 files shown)


TwitchBronBron (approved):

package.json
```diff
@@ -853,6 +853,11 @@
                                             "description": "Should this component library be installed onto the Roku device during the sideloading process? Component libraries are installed in the order they're found in this array, so it is recommended to keep all `install: true` component libraries at the top of the array. \n\nTo utilize this component library in your application, set `dcl_channel_ids=DCLHelloLib:devlib_dcl_1` (where 1 is the 1-based position in this array)",
                                             "default": false
                                         },
+                                        "enablePostfix": {
+                                            "type": "boolean",
+                                            "description": "Should this component library's `.brs` files be renamed with a `__lib<index>` postfix during staging? Postfixing lets the debugger map device-reported file paths back to the component library they came from. Set to `false` to leave file names untouched (for example, when the library loads files by a fixed name at runtime); source mapping for this library's files will be degraded while debugging.",
```
TwitchBronBron: not only brs, we also do .xml, and I believe also any file imported by <script tags which could be any file extension. just need this description more updated.

chrisdp: Good catch. Updated the description to drop the `.brs`-only framing: it now says postfixing renames the library's files and rewrites the references between them, covering BrightScript and XML files plus any file imported via `<script>` tags (which can have any extension).

PR #839, opened 2026-06-30
Reuse probed device-info for launch and host selection

chrisdp:
`promptForHost` (including manual host entry) now returns `{ host, deviceInfo }`, using the raw device-info gathered while probing the device. All callers read `.host`.

`DebugConfigurationProvider` attaches that raw device-info to the launch config so the debug session can skip its own `device-info` request, and enhances a local copy via roku-deploy's `enhanceDeviceInfo` instead of issuing a second request.

Depends on rokucommunity/roku-deploy#300 (`enhanceDeviceInfo`) and rokucommunity/roku-debug#381 (the launch config `deviceInfo` field).


TwitchBronBron (approved):
Looks good to me. :)

PR #855, opened 2026-07-10
Sg inspector remove appui keypaths

triwav:
With the upgrade to RTA 3 we no longer need appUI key paths in the SceneGraph Inspector. This makes the initial request a small amount slower but substantially improves all future user actions that would have required converting the appUI key path into an elementID key path.


TwitchBronBron (approved):

TwitchBronBron:
closing in favor of https://github.com/rokucommunity/vscode-brightscript-language/pull/881 which fully includes these changes (and some more...)

PR #865, opened 2026-07-30
DeviceManager rewrite: order-based discovery

Christian-Holbrook:
## Summary

Implements the behavior described in [`docs/device-discovery.md`](docs/device-discovery.md): the `scanNeeded` flag and direct `refresh()`/`scan()` calls are replaced by an **orders** system where triggers submit work and visible views fulfill it. No new files — this is a behavior change in the existing ones.

> **Note:** the doc will be updated separately to match a few places where the implementation evolved past its wording (called out below).

## How it works now

**Triggers submit orders; views fulfill them.**

```
trigger (startup / network / sleep / config change / refresh click / stale timer)
   └─ submitOrder(reason)          ── one funnel; the reason decides which order
        ├─ broadcast order            types it implies (most imply both;
        └─ reconcile order            config-changed → reconcile only,
                                      unhealthy-device → broadcast only)
                │
                ├─ emits *-ordered  → a VISIBLE view fulfills live
                └─ pending sets     → a hidden view fulfills on open
                        │
                        └─ fulfillment coalesces: ALL pending broadcast
                           reasons → one SSDP M-SEARCH scan; ALL pending
                           reconcile reasons → one health-check sweep
```

- **Pending orders are a set of reasons per order type** — different reasons accumulate (`stale` + `sleep` can both be pending), the same reason never queues twice, and one execution satisfies everything taken. When a real reason triggers fulfillment, the whole set clears (the scan satisfies the stale want too).
- **Reasons carry the policy.** A stale-only broadcast is gated (discovery enabled + last scan older than 30 min); any real reason scans now. A reconcile bypasses each device's cache trust window only when `refresh-clicked` is among its reasons.
- **Views are the gate.** If no UI is visible, no network traffic happens — orders queue until a view opens. The tree view fulfills everything on open (including a panel already open at activation) and live orders except `stale`; the quick pick fulfills except `stale` on open/live, with a 7s fallback that fulfills whatever is pending (staleness-gated inside, so it never over-scans).

**Lazy hydration on read.** Reads return immediately; devices that are `unknown` with no cache, or whose cache is older than 8h, get a background resolve (with convergence guards: pending skip, in-flight set, 5-min per-IP retry cooldown). This is what makes `ssdp:alive` "just work", and it replaces the deleted eager paths (`resolveUncachedDiscoveredDevices`, the post-scan `healthCheckStaleDevices` sweep). Because reads now cost network, the tree view no longer reads while hidden — changes set a dirty flag consumed on the next visible render.

**The de-dupe rule.** Concurrent device-info requests for the same `ip:port` share one in-flight HTTP request; a sequence counter ensures a stale slow response can't overwrite a newer result (regression test included).

**Views speak interactions, not mechanisms.** The full view-facing surface of `DeviceManager`:

| Call | Meaning |
|---|---|
| `submitOrder('refresh-clicked')` | "the user clicked refresh/scan" |
| `fulfillPendingOrders({except})` | "I'm visible — run what's queued" |
| `fulfillPendingBroadcast/Reconcile` | per-order-type view policy (picker's 7s fallback) |
| `deviceEngaged(keyOrDevice)` | "the user engaged this device" — freshens it, reports health |

Everything else (order types, executors, gates, stale timers, the cache-bypass mechanics) is private. The old `force` flag is gone — `resolveDevice` takes `{ bypassCache, syntheticDelay }` and the reason→policy mapping lives inside the manager.

## Deliberate deviations from the doc (to be folded into its next revision)

- **7s picker fallback *fulfills* rather than *submits*** a stale order — per the doc's own view rules, nobody would fulfill a submitted stale order, so it would sit forever.
- **Pending orders are a set of reasons, not a single slot** — a stale order can no longer downgrade (or be lost behind) a queued real trigger.
- **`unhealthy-device` broadcasts are rate-limited** (suppressed when discovery is off or a scan ran within 60s) — terminates the scan → fail → scan loop the reason list would otherwise allow.
- **The discovery-enabled toggle submits `startup` orders** — the doc models no toggle reason.

## Bug fixes along the way

- `randomDelay(min, max)` produced 0–max instead of min–max
- `refreshDevice` passed the encoded tree key as a serial number — silent no-op
- The quick pick's scan-fallback timer leaked when the picker closed early
- The tree view missed queued startup orders when the panel was already open at activation

### DIFF
--- a/src/deviceDiscovery/DeviceManager.ts
+++ b/src/deviceDiscovery/DeviceManager.ts
@@ -666,7 +666,40 @@ export class DeviceManager {
         this.reconcile(reasons);
         return true;
     }
+
+    /**
+     * Fulfill both pending order types in one call — the common case for views (on open, on
+     * becoming visible). Views that need per-order-type policy (e.g. the quick pick's 7s
+     * fallback only wants broadcasts) can still call the specific fulfillments directly.
+     */
+    public fulfillPendingOrders(options?: { except?: Array<BroadcastReason | ReconcileReason> }): { scanStarted: boolean; reconciled: boolean } {
+        return {
+            scanStarted: this.fulfillPendingBroadcast({ except: options?.except as BroadcastReason[] }),
+            reconciled: this.fulfillPendingReconcile({ except: options?.except as ReconcileReason[] })
+        };
+    }
+
+    /**
+     * The "user clicked refresh / scan" interaction (spec: Clicking refresh): always submits a
+     * broadcast order AND a reconcile order, regardless of how recently either ran. Views report
+     * the interaction; which order types it implies is the manager's business.
+     */
+    public submitRefreshOrders(): void {
+        this.submitBroadcast('refresh-clicked');
+        this.submitReconcile('refresh-clicked');
+    }
     // #endregion
+
+    /**
+     * The user engaged with a specific device — clicked it, expanded it, selected it in the
+     * picker, or is about to launch to it (spec: "explicit engagement with that specific
+     * device"). Freshens the device now (bypassing its cache trust window) and reports whether
+     * it responded. Fire-and-forget callers can ignore the result; callers that need to know
+     * (picker validation, pre-launch checks) await it.
+     */
+    public async deviceEngaged(deviceOrLookup: RokuDevice | { ip?: string; serialNumber?: string }): Promise<boolean> {
+        return this.healthCheckDevice(deviceOrLookup, true, false);
+    }
 
     /**
      * Broadcast an SSDP M-SEARCH to discover devices on the network. Does NOT health-check
@@ -752,7 +785,7 @@ export class DeviceManager {
         this.clearCurrentDeviceList();
     }
 
-    public async healthCheckDevice(deviceOrLookup: RokuDevice | { ip?: string; serialNumber?: string }, force = false, doSyntheticDelay = true): Promise<boolean> {
+    private async healthCheckDevice(deviceOrLookup: RokuDevice | { ip?: string; serialNumber?: string }, force = false, doSyntheticDelay = true): Promise<boolean> {
         // If already a device object with deviceState, use it directly; otherwise look it up
         const device = 'deviceState' in deviceOrLookup
             ? deviceOrLookup
--- a/src/deviceDiscovery/DeviceManager.ts
+++ b/src/deviceDiscovery/DeviceManager.ts
@@ -39,8 +39,9 @@ export class DeviceManager {
             //if the `deviceDiscovery.enabled` setting was changed, start or stop monitoring
             if (event?.affectsConfiguration('brightscript.deviceDiscovery.enabled')) {
                 if (this.deviceDiscoveryEnabled) {
-                    //emit that we need a scan (will trigger UI to refresh and show devices as needed when enabled)
-                    this.setScanNeeded(true);
+                    //order a broadcast + reconcile so the views discover/health-check once enabled
+                    this.submitBroadcast('startup');
+                    this.submitReconcile('startup');
                     this.systemSleepMonitor.start();
                     void this.activateMonitoring();
                 } else {
@@ -54,11 +55,11 @@ export class DeviceManager {
                 this.emitDevicesChanged();
             }
 
-            //if the `devices` setting was changed, re-apply configured devices and health check them
+            //if the `devices` setting was changed, re-apply configured devices immediately (cheap,
+            //local) and order a health check for the views to fulfill when one is visible
             if (event?.affectsConfiguration('brightscript.devices')) {
-                this.loadConfiguredDevices().then(() => {
-                    return this.healthCheckAllDevices(false, true);
-                }).catch(() => { });
+                this.loadConfiguredDevices().catch(() => { });
+                this.submitReconcile('config-changed');
             }
 
             //if the `defaultDevicePassword` setting was changed, refresh any device views that rely on it
@@ -86,7 +87,9 @@ export class DeviceManager {
 
     private setupMonitors() {
         this.systemSleepMonitor = new SystemSleepMonitor(() => {
-            this.setScanNeeded();
+            //order a broadcast + reconcile so the views rescan/health-check on wake
+            this.submitBroadcast('sleep');
+            this.submitReconcile('sleep');
         });
         this.networkChangeMonitor = new NetworkChangeMonitor(() => {
             this.networkId = getNetworkHash();
@@ -107,8 +110,9 @@ export class DeviceManager {
 
             this.restartRokuFinder();
 
-            //this is important for telling the devices view to refresh and health check its devices
-            this.setScanNeeded();
+            //order a broadcast + reconcile so the views rescan/health-check on the new network
+            this.submitBroadcast('network');
+            this.submitReconcile('network');
         });
     }
 
@@ -130,14 +134,13 @@ export class DeviceManager {
             // Sleep monitor runs all the time when enabled (ignores focus state)
             this.systemSleepMonitor.start();
 
-            this.activateMonitoring().then(() => {
-                const lastSeenDeviceIds = this.globalStateManager.getLastSeenDevices(this.networkId);
-                if (lastSeenDeviceIds.length === 0) {
-                    this.refresh();
-                } else {
-                    this.setScanNeeded();
-                }
-            }).catch((e) => {
+            //order a broadcast + reconcile for the views to fulfill when they open.
+            //No proactive scan here even on a cold cache — per the design doc, no network
+            //traffic happens until a view is actually visible to consume these orders.
+            this.submitBroadcast('startup');
+            this.submitReconcile('startup');
+
+            this.activateMonitoring().catch((e) => {
                 console.error(e);
             });
         }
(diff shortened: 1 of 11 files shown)


src/deviceDiscovery/DeviceManager.ts
```diff
@@ -39,8 +39,8 @@ export class DeviceManager {
             //if the `deviceDiscovery.enabled` setting was changed, start or stop monitoring
             if (event?.affectsConfiguration('brightscript.deviceDiscovery.enabled')) {
                 if (this.deviceDiscoveryEnabled) {
-                    //emit that we need a scan (will trigger UI to refresh and show devices as needed when enabled)
-                    this.setScanNeeded(true);
+                    //order a broadcast + reconcile so the views discover/health-check once enabled
+                    this.submitOrder('startup');
```
TwitchBronBron: this generic `submitOrder` function is confusing semantically. I'd expect to see `this.submitBroadcastOrder()` and `this.submitReconcileOrder()` instead. Because we're not submitting "startup" orders. 

Maybe more like this
```ts
this.submitOrders(
  { type: 'broadcast', source: 'startup'}, 
  { type: 'reconcile', source: 'startup'
})
```

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +586,185 @@ export class DeviceManager {
...
+    /**
+     * Submit a broadcast (SSDP scan) order. Adds the reason to the pending set AND emits
+     * `broadcast-ordered` so a visible view can fulfill it live; a hidden view consumes the
+     * set when it opens. Reasons accumulate independently (e.g. `stale` and `sleep` can both
+     * be pending) but the same reason never queues twice.
+     */
+    private submitBroadcast(reason: BroadcastReason): void {
+        this.pendingBroadcastReasons.add(reason);
+        const order: BroadcastOrder = { reason: reason, timestamp: Date.now() };
+        this.emitter.emit('broadcast-ordered', order);
```
TwitchBronBron: I'm wonder why we don't just keep a queue of orders, instead of these parallel lists of "reasons". Because we would also lose the audit trail of the timestamp (and any other data we might want to keep in an order). 


I'd imagine it would be more like:
```typescript
const order: BroadcastOrder = { reason: reason, timestamp: Date.now() };
this.orders.push(order);
this.emitter.emit('broadcast-ordered', order);
```

And then anyone who needs info about the whole list of orders, it's easy enough to assmble them from the queue of unhandled orderes.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +586,185 @@ export class DeviceManager {
...
+    }
+
+    /**
+     * Submit a reconcile (health-check-all) order. Same set + event semantics as
+     * {@link submitBroadcast}.
+     */
+    private submitReconcile(reason: ReconcileReason): void {
```
TwitchBronBron: can we name these `submitReconcileOrder` and `submitBroadcastOrder`?

or, `orderReconcile` and `orderBroadcast`?

TwitchBronBron (changes requested):

src/deviceDiscovery/DeviceManager.ts
TwitchBronBron: Let's maybe store it like this:

```typescript
orders: {
    'reconcile': {
        type: 'reconcile',
        reasons: ['stale', 'startup']
    },
    'broadcast': {
        type: 'broadcast',
        reasons: ['stale']
    }
}
```

src/deviceDiscovery/DeviceManager.ts
TwitchBronBron: Let's just eliminate these functions, in favor of the this.submitOrders function.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +605,152 @@ export class DeviceManager {
...
+     * reasons by one health-check sweep (cache bypassed when `refresh-clicked` is among them).
+     *
+     * @param options.types - which order types to fulfill. Callers state their policy
+     *   explicitly — e.g. the tree view fulfills both, the quick pick's 7s fallback only wants
+     *   broadcasts, live handlers pass the submitted order's type.
+     * @param options.except - reasons that do not TRIGGER fulfillment on their own, e.g.
+     *   `{ except: ['stale'] }` — see {@link Orders.take} for the full semantics.
+     */
+    public fulfillOrders(options: { types: OrderType[]; except?: Array<BroadcastReason | ReconcileReason> }): { scanStarted: boolean; reconciled: boolean } {
+        const result = { scanStarted: false, reconciled: false };
```
TwitchBronBron: I thnk we can just `return this.orders.take(options)` instead. This is just abstracting things for almost no practical benefit?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +605,152 @@ export class DeviceManager {
...
+     * Atomically consume AND execute pending orders — the one-call fulfillment API for views.
+     * The work is idempotent per order type, so all taken reasons are satisfied by a single
+     * execution: broadcast reasons by one scan (staleness-gated when stale-only), reconcile
+     * reasons by one health-check sweep (cache bypassed when `refresh-clicked` is among them).
+     *
+     * @param options.types - which order types to fulfill. Callers state their policy
+     *   explicitly — e.g. the tree view fulfills both, the quick pick's 7s fallback only wants
+     *   broadcasts, live handlers pass the submitted order's type.
+     * @param options.except - reasons that do not TRIGGER fulfillment on their own, e.g.
+     *   `{ except: ['stale'] }` — see {@link Orders.take} for the full semantics.
```
TwitchBronBron: Let's remove all the emdash chars in this file (and maybe tighten up the comments too)

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +605,152 @@ export class DeviceManager {
...
-     * Trigger a network scan for devices without health checking existing devices.
-     * Use this when you just want to discover new devices without verifying existing ones.
-     * @param force - If true, scan even if deviceDiscovery is disabled
-     * @returns true if a scan was started, false otherwise
+     * Health-check a single device NOW: bypass its cache trust window, fetch fresh device
+     * info from the network, update the cache (so every view renders the freshest data), and
+     * report whether it responded. Fire-and-forget callers can ignore the result — the cache
+     * update is the point. Accepts anything {@link getDevice} can look up (encoded tree key,
+     * ip/serial lookup) or a device object; an unknown device is simply reported unhealthy.
      */
```
TwitchBronBron: I don't think we need to specify how thsi works under the hood. this function's job is to "ensure device still exists on the network". That's it. The fact that it runs deviceInfo fetch underneath isn't important for the consumer.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +605,152 @@ export class DeviceManager {
...
+     * {@link healthCheckDevice} (bypass cache trust window, fetch, update cache) — this variant
+     * returns the resulting device info for callers that need it (e.g. pre-launch host
+     * resolution). Returns undefined when the device is unknown or unreachable.
+     */
+    public async getDeviceInfo(deviceKeyOrLookup: RokuDevice | string | { ip?: string; serialNumber?: string }): Promise<Record<string, any> | undefined> {
+        return (await this.fetchFreshDevice(deviceKeyOrLookup))?.deviceInfo;
+    }
+
+    /**
+     * The shared engine behind {@link healthCheckDevice} and {@link getDeviceInfo}: resolve
```
TwitchBronBron: Unnecessary "shared engine" bit of the comment here, we can already find out who calls it, why mention it again?

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -504,36 +605,152 @@ export class DeviceManager {
...
+    /**
+     * Health-check every known device (configured + discovered), marking them online/offline.
+     * Does NOT scan for new devices — that's {@link broadcast}. The reasons decide the urgency:
+     * only an explicit user click bypasses each device's cache trust window — a bypassing
+     * reconcile is one HTTP request per known device, so only "I want fresh data NOW" earns
+     * that. Reachable only through order fulfillment.
+     */
+    private reconcile(reasons: ReconcileReason[]): void {
+        const bypassDeviceCache = reasons.includes('refresh-clicked');
+        this.healthCheckAllDevices(bypassDeviceCache).catch(() => { });
```
TwitchBronBron: Should we be eating the error here? At least should log it, even if there's no recovery.

src/deviceDiscovery/DeviceManager.ts
```diff
@@ -1049,65 +1273,51 @@ export class DeviceManager {
...
-        await Promise.all([...staleIps].map(ip => this.resolveDevice({ ip: ip }, false)));
+        return inFlight;
     }
 
     /**
      * Fetch device info from the network. Always makes a network request.
-     * Caches the result in globalStateManager for future lookups.
      */
-    private async fetchDeviceInfo(ip: string, port: number): Promise<DeviceInfoRaw> {
+    private async fetchDeviceInfoInner(ip: string, port: number): Promise<DeviceInfoRaw> {
```
TwitchBronBron: Let's call _this one_ `getDeviceInfo`. It never returns cached data. It always hits the device directly. it should be public. 

If you want cache help, you'll call `.getDevice()` instead. 

This function can also be the one that's throttled (eliminating the `fetchDeviceInfo` call entirely). You could lean into some type of method annotation like `@throttle()` too. (likely want it to be the reverse though...the first call gets activated, the others piggback on the call and get the same answer back).

TwitchBronBron (changes requested):

PR #878, opened 2026-08-13
[fix] resolve brightscript.bsdk correctly when .code-workspace is not at project root

addison-adler:
### Context

- `getConfiguration(...).get('bsdk')` pre-resolves relative paths and variables like `${workspaceFolder}` against the wrong base when the `.code-workspace` file is not at the project root
  - `${workspaceFolder}/node_modules/brighterscript` gets resolved to `node_modules/brighterscript`
- then, `LanguageServerManager.ts` uses the workspace file as the `cwd` location

net result:
- `.code-workspace` doesn't work anywhere except for the project root

### Changes

Updated behavior:
- if `brightscript.bsdk` starts with `${workspaceFolder}`, resolve relative to the first workspace folder (same behavior as VSCode)
- if `brightscript.bsdk` starts with `${workspaceFolder:name}`, resolve relative to the named workspace folder
- **else, same behavior as previously!** 

## How I Tested

- built the extension locally,
- tested `${workspaceFolder}`, `${workspaceFolder}`, and plain string path
- tested both relative and absolute paths

## Test plan

- [ ] Place `.code-workspace` file in a subdirectory (e.g. `.vscode/`) with a relative `brightscript.bsdk` path — language server should resolve correctly
- [ ] Use `${workspaceFolder:FolderName}/node_modules/brighterscript` — should resolve to the named folder's path
- [ ] Use an unrecognized variable like `${env:MY_VAR}` — should surface a clear error in the BrightScript output panel
- [ ] Single-root workspace (no `.code-workspace` file) — behavior unchanged


src/LanguageServerManager.ts
```diff
@@ -557,6 +558,55 @@ export class LanguageServerManager {
...
+        }
+
+        const [match, workspaceName, relativePath] = /^\$\{workspaceFolder:?([^}]*)\}(.*)$/.exec(value) ?? [];
+
+        if (!match) {
+            throw new Error(`brightscript.bsdk: unsupported variable in bsdk "${value}"`);
+        }
+
+        // if ${workspaceFolder}, use workspaceFolders[0]
+        let workspaceFolder = vscode.workspace.workspaceFolders?.[0]
```
addison-adler:
```suggestion
        let workspaceFolder = vscode.workspace.workspaceFolders?.[0];
```

src/LanguageServerManager.ts
```diff
@@ -557,6 +558,55 @@ export class LanguageServerManager {
...
+        // if ${workspaceFolder:name}, find by name
+        if (workspaceName) {
+            workspaceFolder = vscode.workspace.workspaceFolders?.find(f => f.name === workspaceName);
+        }
+
+        if (!workspaceFolder) {
+            throw new Error(`brightscript.bsdk: unknown workspace folder name "${workspaceName}"`);
+        }
+
+        const bsdkInfo = path.join(workspaceFolder.uri.fsPath, relativePath)
```
addison-adler:
```suggestion
        const bsdkInfo = path.join(workspaceFolder.uri.fsPath, relativePath);
```

src/LanguageServerManager.spec.ts
```diff
@@ -270,6 +270,97 @@ describe('LanguageServerManager', () => {
...
+            // .code-workspace file lives in a subdirectory (.vscode/)
+            vscode.workspace.workspaceFile = URI.file(s`${tempDir}/.vscode/workspace.code-workspace`);
+            vscode.workspace.workspaceFolders.push({
+                index: 0,
+                name: 'SDK',
+                uri: URI.file(s`${tempDir}`)
+            });
+
+            setConfig(vscode.workspace.workspaceFile.fsPath, {
+                'brightscript.bsdk': '../node_modules/brighterscript'
```
TwitchBronBron: I think this will result in a regression. Currently in the extension, if I run the "change selected brighterscript version" and then pick my node_modules version, it populates settings.json with this:
```json
{
    "brightscript.bsdk": "node_modules/brighterscript"
}
```

That's equivalent to `./node_modules/brighterscript`, which is _not_ relative to the `.vscode/` folder, it's relative to the workspace root. 

This is the same behavior that typescript implements, so I'm not sure this specific relative path logic should be changed (and it would be a breaking change).

src/LanguageServerManager.ts
```diff
@@ -524,17 +524,18 @@ export class LanguageServerManager {
         //use bsdk entry in the code-workspace file
         if (this.workspaceConfigIncludesBsdkKey()) {
             let result = this.parseVersionInfo(
-                util.getConfiguration('brightscript', vscode.workspace.workspaceFile).get<string>('bsdk')?.trim?.(),
+                this.getWorkspaceBsdkInfo(vscode.workspace.workspaceFile),
```
TwitchBronBron, line 527: Not a fan of mixing workspaceFile and per-folder logic into one function. Should be:

- collect each folder's settings value
- collect the code-workspace value
- if code-workspace value exists, use it; else fall back to folder values

Right now that's inter-mixed inside getWorkspaceBsdkInfo's folder loop, so it's blurry (and weird, because in this case I think every folder's value would be identical since we're defaulting to the code-workspace's value). I'd expect the .code-workspace handling to live as separate logic after the per-folder loop, not merged into it.

Something more like this:

Image

TwitchBronBron (changes requested):
I like the `${workspaceFolder}` and `${workspaceFolder:name}` concept. But I think the current logic needs a little bit of refactoring to keep the logic a little easier to follow.

